PR #12, opened 2026-03-22
Planning dyn2d

Jaraxxus-Me:
- Fixed a skill definition bug in dynobstruction2d
- Updated the bilevel planning model to be consistent with kinematic obstruction2d
- Verified that dynobstruction-o0 could succeed sometimes (but not all times, as the target block is not necessarily graspable).
- DynObstruction2d-o1 should be ready for experiments now.


yixuanhuang98 (approved):
LGTM!

PR #14, opened 2026-03-23
Vlm predicates

Jaraxxus-Me:
An initial implementation of querying a VLM for state_abstractor in bilevel planning baselines.


yixuanhuang98 (approved):
Approving to unblock.

PR #29, opened 2026-04-15
Changed dynamic3d test task location

vaibhavsaxena11:
Made tests refer to local task json files for dynamic3d tasks.


yixuanhuang98 (commented):
I am not very sure about the changes in this PR. seems like we updated env = kinder.make -> env = TidyBot3DEnv. 

This would automatically make the unit tests work only with the specific configs, instead of supporting all the variants we had in Kindergarden.

vaibhavsaxena11:
> I am not very sure about the changes in this PR. seems like we updated env = kinder.make -> env = TidyBot3DEnv.
> 
> This would automatically make the unit tests work only with the specific configs, instead of supporting all the variants we had in Kindergarden.

To reduce clutter in the `tasks` folder in dynamic3d envs in kindergarden, I moved all test json files to the test folders. This made sure that the task jsons in `kindergarden/envs/dynamic3d/tasks` solely contain the tasks that should be registered, and no test tasks.

Hence, to access the test tasks, the easiest way is to currently refer to them directly using their path the tests folder. I have followed this approach in kindergarden already, and am mimicing the same here for baselines. Let me know what you think

yixuanhuang98:
> > I am not very sure about the changes in this PR. seems like we updated env = kinder.make -> env = TidyBot3DEnv.
> > This would automatically make the unit tests work only with the specific configs, instead of supporting all the variants we had in Kindergarden.
> 
> To reduce clutter in the `tasks` folder in dynamic3d envs in kindergarden, I moved all test json files to the test folders. This made sure that the task jsons in `kindergarden/envs/dynamic3d/tasks` solely contain the tasks that should be registered, and no test tasks.
> 
> Hence, to access the test tasks, the easiest way is to currently refer to them directly using their path the tests folder. I have followed this approach in kindergarden already, and am mimicing the same here for baselines. Let me know what you think

Thanks for the explanation! Due to the changes in KinDERGarden, did we reduce the number of variants per environment?

vaibhavsaxena11:
> > > I am not very sure about the changes in this PR. seems like we updated env = kinder.make -> env = TidyBot3DEnv.
> > > This would automatically make the unit tests work only with the specific configs, instead of supporting all the variants we had in Kindergarden.
> > 
> > 
> > To reduce clutter in the `tasks` folder in dynamic3d envs in kindergarden, I moved all test json files to the test folders. This made sure that the task jsons in `kindergarden/envs/dynamic3d/tasks` solely contain the tasks that should be registered, and no test tasks.
> > Hence, to access the test tasks, the easiest way is to currently refer to them directly using their path the tests folder. I have followed this approach in kindergarden already, and am mimicing the same here for baselines. Let me know what you think
> 
> Thanks for the explanation! Due to the changes in KinDERGarden, did we reduce the number of variants per environment?

No, none of the variants were removed. See [this PR](https://github.com/Princeton-Robot-Planning-and-Learning/kindergarden/pull/59).

tomsilver (approved):
LGTM thanks!

PR #31, opened 2026-04-20
Update run commands for VLM baselines

vaibhavsaxena11:


yichao-liang (approved):
Looking good, thanks! My only suggestion is that we probably don’t need to enumerate essentially the same experiment commands for all eight environments?

PR #34, opened 2026-04-22
set default scene_bg = true

yixuanhuang98:
set default scene_bg = true for dynamic3D environments in both kinder-models and kinder-bilevel-planning


vaibhavsaxena11:
Won't this need downloading mimiclabs assets in CI?

PR #84, opened 2026-07-26
sweep3D skills: resample on infeasible sample instead of crashing

LJ1356:
## Problem

The sweep3D parameterized controllers (`open_drawer`, `pick_wiper`, `sweep`) raise an **uncaught** exception from `reset()` whenever a sampled parameter is infeasible:

- `inverse_kinematics(...)` raises `InverseKinematicsError`;
- `run_base_motion_planning(...)` / `run_motion_planning(...)` return `None` and hit a bare `assert`.

These propagate out of the trajectory sampler and abort the entire planning episode. Because `BacktrackingRefiner` only catches `TrajectorySamplingFailure`, it cannot retry with a different sample — so the `samples_per_step` budget is effectively wasted the moment a single sample is infeasible.

## Fix

Convert those failure points to `TrajectorySamplingFailure`, which the refiner catches and resamples:

- a small `_ik_or_resample(...)` helper wraps the IK calls;
- `None` motion plans raise `TrajectorySamplingFailure` explicitly instead of asserting;
- the sweep's `smoothly_follow_end_effector_path(...)` is wrapped to translate `InverseKinematicsError`.

**No behavior change when samples are feasible** — the asserts never fired in that case anyway. The change is confined to `parameterized_skills.py` (1 file, +47/-24).

## Motivation

This makes backtracking refinement robust to infeasible parameter samples, which is a prerequisite for exploring non-degenerate parameter sampling bounds (currently the sweep3D bounds are single points, so infeasible samples never arise — but any widening immediately trips the crash).

### DIFF
--- a/kinder-models/src/kinder_models/dynamic3d/sweep3D/parameterized_skills.py
+++ b/kinder-models/src/kinder_models/dynamic3d/sweep3D/parameterized_skills.py
@@ -7,6 +7,9 @@ from bilevel_planning.structs import (
     GroundParameterizedController,
     LiftedParameterizedController,
 )
+from bilevel_planning.trajectory_samplers.trajectory_sampler import (
+    TrajectorySamplingFailure,
+)
 from gymnasium.spaces import Box
 from kinder.envs.dynamic3d.object_types import (
     MujocoDrawerObjectType,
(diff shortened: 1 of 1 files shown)


yixuanhuang98 (approved):
LGTM!

PR #85, opened 2026-07-28
Transport3D: fix arm-plan re-timing metric and place-retract planning (0.80 -> 0.967)

Alpslee:
## Summary

Two mechanism-backed fixes to the Transport3D DS-policy skill stack. Both are gated behind instance attributes that only the Transport3D policy sets, so the other consumer of `base_controllers.py` (Shelf3D, already at 1.00) is byte-for-byte unchanged.

### 1. Arm plans were re-timed with the wrong metric

`remap_joint_position_plan_to_constant_distance` was called with `max_distance = max_action_mag / 2 = 0.1` under the default weighted-L1 joint metric, but the env clips each joint **independently** at `|dq_i| <= max_action_mag = 0.2`. With several joints moving together the step budget is 4-6x tighter than the env allows: measured mean commanded `max|dq|` was 0.0504 against an allowance of 0.20 — 75% of the action authority unused on every step. That is what pushed successful episodes to 1085-1327 steps against the 1200-step cap.

Fix: re-time with an L-infinity, circular-joint-aware distance function at `0.9 * max_action_mag` (opt-in via `_use_linf_arm_remap`). Successful episodes drop to 240-346 steps. Monotonicity control: the same metric at 0.03 makes 0/8 episodes finish in time.

### 2. A place retract could not plan from its own start state

After releasing, the gripper is still inside the box, so the retract planner counts the start configuration as in-contact and returns `None` — even though the env accepted every action that produced that state. The resample loop then re-grounds `place` with nothing grasped and crashes on `assert grasped_object_transform is not None` (12.8% of episodes).

Fix: on a `None` retract plan, retry once with the bodies currently touching the robot subtracted from the collision set — the pick retract already does the analogous thing via `_get_inside_object_ids()`, which returns `[]` for a place. Bounded to once per controller instance (unbounded, the resample ladder triggers an extra motion-planning call per candidate and episodes stall).

### DS policy changes

Cube `(pick, place)` pairs execute greedily with deferral of currently-unreachable cubes (the box move stays last), approach parameters resample on planning failure, and `place` walks a deterministic `(distance, rot)` approach ladder whose index 0 reproduces the original hard-coded `(0.65, 0.0)` exactly. `T3D_BASELINE=1` restores the original single-attempt behavior for A/B comparison.

## Results

Paired protocol eval (4 seeds x 10 episodes, `success = terminated`):

| | baseline | this PR |
|---|---|---|
| Transport3D-o2 | 0.80 | **29/30 = 0.967** |

Per seed: 10/10, 10/10 (seeds 0, 3), 5/6 (seed 1), 4/4 (seed 2; last episodes still running when tallied).

## Tests

- `kinder-ds-policies/tests/policies/kinematic3d/test_transport3d.py`: 10 passed (new coverage for pair scheduling, deferral, and the approach ladder)
- `kinder-models/tests/kinematic3d/`: 12 passed, including shelf3d against the shared `base_controllers.py`

### DIFF
--- a/kinder-ds-policies/src/kinder_ds_policies/policies/kinematic3d/transport3d.py
+++ b/kinder-ds-policies/src/kinder_ds_policies/policies/kinematic3d/transport3d.py
@@ -5,6 +5,7 @@ parameterized skills from kinder-models. The logic mirrors the test in kinder-
 models/tests/kinematic3d/transport3d.
 """
 
+import os
 from typing import Any
 
 import numpy as np
@@ -27,6 +28,23 @@ from kinder_ds_policies.policies.base import PolicyFailure, StatefulPolicy
 
 __all__ = ["create_domain_specific_policy"]
 
+_SkillSpec = tuple[str, tuple[Object, ...], NDArray[np.float32]]
+_SkillPair = list[_SkillSpec]
+
+
+def _place_approach_candidates() -> list[tuple[float, float]]:
+    """(stand-off distance, approach rotation) candidates for the place skill.
+
+    Ordered by |rot| then by distance so that index 0 is exactly the skill's original
+    hard-coded (0.65, 0.0) approach: if that works nothing changes. Rotating the park
+    position around the target surface is what makes a blocked approach solvable -- the
+    robot always faces the target, so the place itself stays valid.
+    """
+    rots = [0.0]
+    for k in range(1, 7):
+        rots.extend([k * np.pi / 12, -k * np.pi / 12])
+    return [(d, r) for r in rots for d in (0.65, 0.55, 0.72)]
+
 
 class Transport3DScriptedPolicy(StatefulPolicy):
     """A stateful scripted policy for Transport3D.
@@ -34,6 +52,9 @@ class Transport3DScriptedPolicy(StatefulPolicy):
     This policy maintains state between calls to track which controller is currently
     executing and the overall progress through the task.
     """
+
+    #: (distance, rot) base-approach candidates for `place`, |rot|-ascending.
+    PLACE_APPROACHES = _place_approach_candidates()
 
     def __init__(
         self,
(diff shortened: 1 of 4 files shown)


tomsilver (commented):
Thanks for this PR! I read it closely and the two mechanisms check out: the env really does clip each joint independently, and BiRRT really does reject an in-contact start configuration before doing anything else. Nice detective work, and the gating means Shelf3D is untouched as claimed. I'd like to merge this.

A few requests first. If you don't have bandwidth for them, just say so and I'll take care of them in a follow-up — the contribution is very welcome either way.

1. Can you share the exact commands or script you used for the eval in the description, so I can reproduce the 29/30 locally? I'd also like the final tally for seeds 1 and 2, which were still running when you posted.

2. `_bodies_touching_robot` reads the PyBullet contact cache without calling `performCollisionDetection`, so it returns contacts for whatever configuration the planner checked last. That happens to be the start configuration in the failure mode you're targeting (BiRRT checks the start first and short-circuits), but not when the goal is in collision or iterations run out, in which case the rescue subtracts arbitrary bodies. Re-setting the sim to `self._current_state` and calling `p.performCollisionDetection` before reading contacts would make it correct in all cases.

3. `_resample_params` samples outside the pick skill's declared params Box (distance 0.45-0.62 vs (0.5, 0.6), rot ±π vs ±π/4), and the place approach `(distance, rot)` is injected via setattr rather than through the parameter vector. Since these policies generate demonstrations, I want controller behavior to stay a function of (skill, params); otherwise two demos with the same recorded params can execute differently. Widening the declared bounds and making the approach a real place parameter would fix this.

4. Small one: the grasp-stall change in `GroundPickController` isn't behind a gate, and those skills are also imported by kinder-bilevel-planning's transport3d env models. I think fail-fast is right there too, but please note it in the PR description since it also changes what `T3D_BASELINE=1` measures.

Thanks again!

PR #87, opened 2026-08-05
kinder-models: disconnect PyBulletSim's PyBullet client when the sim is garbage collected

joshnroy:
**TL;DR** — `PyBulletSim` opens a PyBullet client in its constructor and only releases it in `close()`, which has **zero callers** in either [kinder-baselines](https://github.com/Princeton-Robot-Planning-and-Learning/kinder-baselines) or [kindergarden](https://github.com/Princeton-Robot-Planning-and-Learning/kindergarden). Since `ground()` returns a fresh controller per call and each builds a sim lazily, **one skill execution leaks one client and ~150 MB, without bound** — measured at +150.0 MB/execution, R² = 0.999974, live clients exactly `iter + 1` on 40/40 iterations, no plateau. It reproduces in this repo's own `test_pick_toss`. Fixed by registering a `weakref.finalize` that disconnects the client when the sim is collected: **2 files, +103/−2**, with a regression test that fails on `main` and passes with the fix.

image

Reported in #88.

## Question / goal

Why does any loop that repeatedly executes `dynamic3d/tossing` skills grow in memory until it is OOM-killed, and what is the smallest change that fixes it?

## Hypothesis

`PyBulletSim.__init__` connects a client that only `close()` releases; `close()` is never called and there is no finalizer, so one client leaks per skill execution. If so, live client count and RSS should grow linearly and without bound, and explicitly disconnecting should flatten both.

## Guidance given

The fix was constrained to be **the minimum that closes the leak** while following this repo's existing conventions — no refactors, no new public API, nothing that is "while we're here". Assertions had to be on live client count rather than RSS, since the count is exact and deterministic while RSS is allocator-dependent. The regression test had to be written first and observed failing before the fix existed.

## Methods

`PyBulletSim.__init__` opens a `DIRECT` client (`dynamic3d/utils.py:432`; GUI branch at `:428`) and only releases it in `close()` (`:590`). `close()` has **zero callers** anywhere in [kinder-baselines](https://github.com/Princeton-Robot-Planning-and-Learning/kinder-baselines) or [kindergarden](https://github.com/Princeton-Robot-Planning-and-Learning/kindergarden), and the class defines no finalizer. (The two `sim.close()` calls that a grep turns up in kindergarden are on `ObjectCentricTransport3DEnv` and `ObjectCentricPrplLab3DEnv`; kindergarden does not import `kinder_models` at all, so it cannot reach this class.)

The tossing controllers build a sim lazily in `reset()` (`tossing/parameterized_skills.py:208`, `:335`, `:489`), and `LiftedParameterizedController.ground()` returns a fresh controller on every call:

```python
# bilevel_planning/structs.py:138-145
return self.controller_cls(objects)
```

So one skill execution → one controller → one client → never disconnected.

Four arms were measured, 40 skill executions each, one process per arm, sampling live clients via `p.getConnectionInfo` and RSS from `/proc/self/status`. All four ran inside an 8 GB memory-capped cgroup, since the leak is unbounded by construction.

## Results

| arm | live clients, first → last | RSS growth | RSS slope |
| --- | --- | --- | --- |
| unpatched, fresh controller per execution | 1 → 40 | +5845.0 MB | +150.0 MB/exec, R² = 0.999974 |
| **patched (this PR)** | **0 → 0** | −24.2 MB | +0.1 MB/exec |
| control: one controller reused | 1 → 1 | +0.1 MB | +0.0 MB/exec |
| control: explicit `close()` each execution | 0 → 0 | −17.0 MB | +0.2 MB/exec |

All four arms executed the same 620 environment steps and every iteration reached `terminated()`, so this is not an early-exit artifact. The explicit-`close()` control recovers the full ~150 MB per execution, which is what identifies the un-disconnected client as the whole of the growth.

**It reproduces in this repo's own test suite.** Calling `tests/dynamic3d/tossing/test_tossing_parameterized_skills.py::test_pick_toss` three times in one process:

```
before:                          rss=260.5 MB   live_clients=0
after test_pick_toss() call 1/3: rss=1085.7 MB  live_clients=4
after test_pick_toss() call 2/3: rss=1699.1 MB  live_clients=8
after test_pick_toss() call 3/3: rss=2296.5 MB  live_clients=12
```

That test calls `env.close()`, which releases nothing — the clients belong to the controllers, not the env.

**The change** is two files:

- **`dynamic3d/utils.py`** (+15/−2) — register a `weakref.finalize` that disconnects the client when the sim is collected, and route `close()` through it.
- **`tests/dynamic3d/tossing/.../test_tossing_parameterized_skills.py`** (+88) — a regression test asserting the live client count does not grow across repeated groundings.

Two details are load-bearing and easy to get wrong:

- **The callback takes the client id by value and never closes over `self`.** A bound method such as `weakref.finalize(self, self.close)` would hold a strong reference, the sim would never become unreachable, and the finalizer would never run — the leak would *appear* fixed and would not be. Verified by confirming the sim is collected by refcounting alone with `gc` disabled. This is the same shape `tempfile.TemporaryDirectory` uses in the standard library, which registers a classmethod plus the path by value for exactly this reason; a classmethod is unnecessary here because the cleanup is a single module-level call.
- **Note that garbage collection runs the registered callback, not `close()` itself.** `close()`'s docstring spells out what runs in each case, so that anything added to it later is not silently skipped on collection. A `__del__` calling `close()` would invoke the method, but the CPython docs favour finalizers, and `kindergarden`'s own `mujoco_utils.py:993` is the cautionary example — its unguarded `__del__` raises `AttributeError` on a partially-constructed object.
- **`close()` delegates to the finalizer** rather than calling `p.disconnect` directly. `p.disconnect` on an already-disconnected id raises rather than no-oping, and PyBullet recycles client ids by first-free-slot scan over `MAX_PHYSICS_CLIENTS`, so a stale finalizer holding a recycled id could silently disconnect an unrelated simulator. Routing both paths through the finalize object makes them mutually exclusive, since it marks itself dead on first call. `close()`'s docstring records this.

**Validation:**

- The regression test fails on `main` — `AssertionError: Leaked 4 PyBullet clients over 4 skill executions` — and passes with the fix. It reaches the client-count assertion only after asserting 4/4 executions terminated in more than one step, so it cannot pass vacuously.
- `pytest tests/dynamic3d` — **32 passed**, including `test_pick_toss` and `test_pick_ground_toss`.
- `pylint --rcfile=.pylintrc` on the changed files — **10.00/10**. `black --check` clean.
- `gc.collect()` is **not** load-bearing: with `gc` disabled, refcounting alone frees the sim, so a real planner loop releases promptly.

One side effect not claimed above: because the finalizer is registered immediately after `p.connect`, a failure inside `create_pybullet_robot` (`utils.py:451`) now also releases the client. Previously a construction failure leaked it.

## Recommendation

Merge. A finalizer rather than an explicit `close()` call because the controller is dropped by the planner, so there is no lifecycle hook on which to hang one — adding it would mean introducing a controller teardown protocol and threading it through every caller.

**Deliberately out of scope:**

- `dynamic3d/shelf` (`:740`) and `dynamic3d/sweep3D` (`:1067`) accept an optional `pybullet_sim` that `dynamic3d/tossing` does not. Adding it here would reduce how often a client is created but would **not** fix the leak, since nothing disconnects an injected sim either.
- Context-manager support (`__enter__`/`__exit__`) was considered and dropped: none of the 21 `PyBulletSim(...)` construction sites in these two repos uses `with`, so it would be an unused API in a bug-fix change.

### DIFF
--- a/kinder-models/src/kinder_models/dynamic3d/utils.py
+++ b/kinder-models/src/kinder_models/dynamic3d/utils.py
@@ -432,27 +432,9 @@ class PyBulletSim:
         else:
             self._physics_client_id = p.connect(p.DIRECT)
 
-        # Disconnect when this object is garbage collected. Callers that ground a
-        # fresh controller per skill execution create one sim each, and without this
-        # every one of those clients stays open for the life of the process.
-        #
-        #   weakref.finalize(self, p.disconnect, self._physics_client_id)
-        #                     ^     ^             ^
-        #                     |     |             a plain int
-        #                     |     a module-level function
-        #                     the object being watched
-        #
-        # Both arguments are deliberately self-free. Registering a bound method
-        # (weakref.finalize(self, self.close)) would keep self reachable, so the sim
-        # would never be collected, the callback would never fire, and the leak would
-        # look fixed while continuing. See close() for what does and does not run.
-        #
-        # This mirrors tempfile.TemporaryDirectory in the standard library, which
-        # registers a classmethod plus the path by value for the same reason. It uses
-        # a classmethod because its cleanup is several statements; ours is one call,
-        # so the module-level p.disconnect suffices. If more cleanup is ever needed,
-        # move it to a @staticmethod taking the client id and call that from both
-        # here and close() -- do not reach for self.
+        # Disconnect on collection -- callers ground a fresh controller, and so a
+        # fresh sim, per skill execution. The callback must not reference self, or
+        # the sim stays reachable and is never collected. See close().
         self._finalizer = weakref.finalize(self, p.disconnect, self._physics_client_id)
 
         # Create the robot, assuming that it is a kinova gen3.
@@ -614,19 +596,8 @@ class PyBulletSim:
     def close(self) -> None:
         """Close the PyBullet simulator. Safe to call more than once.
 
-        What runs when, given the finalizer registered in __init__:
-
-            sim.close()          -> this body, then p.disconnect(id)
-            sim goes out of scope -> p.disconnect(id) ONLY; this body does not run
-            close() then collected -> nothing the second time
-
-        Note the middle line: garbage collection invokes the registered callback,
-        not this method, so anything added to this body will silently not happen on
-        collection. Register it in __init__ instead -- and keep it off self, or the
-        sim stays reachable and is never collected at all.
-
-        Do not disconnect physics_client_id by hand instead of calling this:
-        PyBullet reuses client ids, so a finalizer left holding a recycled id would
-        disconnect an unrelated simulator.
+        Collection runs the finalizer, not this method, so put any further cleanup in
+        __init__'s callback. Do not call p.disconnect by hand: ids get reused, so a
+        stale finalizer would disconnect an unrelated sim.
         """
         self._finalizer()
--- a/kinder-models/src/kinder_models/dynamic3d/utils.py
+++ b/kinder-models/src/kinder_models/dynamic3d/utils.py
@@ -1,6 +1,7 @@
 """Utils for tidybot environments."""
 
 import math
+import weakref
 from typing import Iterable
 
 import numpy as np
@@ -431,6 +432,12 @@ class PyBulletSim:
         else:
             self._physics_client_id = p.connect(p.DIRECT)
 
+        # Disconnect when this object is garbage collected, since callers that create
+        # one sim per skill execution would otherwise exhaust PyBullet's client pool.
+        # The callback must take the client id by value and never close over self, or
+        # the sim would stay reachable and never be collected.
+        self._finalizer = weakref.finalize(self, p.disconnect, self._physics_client_id)
+
         # Create the robot, assuming that it is a kinova gen3.
         self._robot = create_pybullet_robot(
             "kinova-gen3",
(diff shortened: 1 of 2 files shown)


tomsilver (approved):
Beautiful work! I learned two new things reviewing this PR: (1) `weakref` is a thing (seems perfect for this use case) and (2) "belt and braces" is an (apparently British) idiom :)

I do have one or two questions for my understanding. If we had been more careful about calling `close()`, would that have solved the same problem? With your change, we don't need to call `close()`, but we do need to make sure that we aren't creating more than ~1000 controllers at a time -- and to be safe, we need to use `del`, or else hope that gc is friendly. Is that correct understanding?

In other words, in the code you originally wrote that led to the memory blowup, will you now need to do `del` (instead of `close()`)? Will users still need to be careful?

joshnroy:
> Beautiful work! I learned two new things reviewing this PR: (1) `weakref` is a thing (seems perfect for this use case) and (2) "belt and braces" is an (apparently British) idiom :)
> 
> I do have one or two questions for my understanding. If we had been more careful about calling `close()`, would that have solved the same problem? With your change, we don't need to call `close()`, but we do need to make sure that we aren't creating more than ~1000 controllers at a time -- and to be safe, we need to use `del`, or else hope that gc is friendly. Is that correct understanding?
> 
> In other words, in the code you originally wrote that led to the memory blowup, will you now need to do `del` (instead of `close()`)? Will users still need to be careful?

Thanks for the quick review!
My understanding is:

1. `close()` _would_ have worked, I'll double check that and attach a snippet before merging (though it's [opinion] an antipattern to force the user to close something manually). We could use a python [context manager](https://book.pythontips.com/en/latest/context_managers.html) if that's instead preferred?
2. `del` is only called in the test to force the garbage collector to collect it. I believe that cpython/any reasonable python interpreter should do this as needed for memory management (aka users don't need to do this), but I'll see if I can add a test to assert this.
3. This doesn't erase the problem of creating arbitrarily many environment simultaneously - I don't think there's much we can do about that (since they'd all be valid environments). Unless we added a max environment limit, but that seems like it would get annoying for users

❇️ hand-written comment by Josh Roy, not made with Claude

tomsilver:
Good enough for me! I agree with your opinion. This is a clear improvement. Context managers would get ugly quickly. Let's merge!

PR #96, opened 2026-08-06
Add a CLAUDE.md documenting the repo's conventions

joshnroy:
## Summary

The repo has no `CLAUDE.md`, no `CONTRIBUTING.md` and no PR template, so everything a
new contributor needs is tacit: which commands are the gate, why linting by hand can
pass without actually linting, which packages a change will trigger in CI, what a test
in this repo is supposed to look like, and a handful of `bilevel_planning` behaviours
that are easy to get wrong. This writes that down, derived from the tree rather than
from general advice.

It documents, among other things:

- **The per-package toolchain.** Ten packages, each carrying its own `pyproject.toml`,
  `.pylintrc`, `pylint_plugins/`, `run_autoformat.sh` and (usually) `run_ci_checks.sh`;
  intra-repo dependencies declared in `prpl_requirements.txt` and consumed by
  `scripts/generate_topological_order.py`.
- **The pylint plugin fails open.** `load-plugins=pylint_plugins.no_np_random` resolves
  against `sys.path`. Running pylint by hand without `PYTHONPATH=<package-dir>` prints
  `E0013: Plugin ... is impossible to load`, skips the `np.random` ban, runs every
  other checker, and still reports `10.00/10`. Verified both ways.
- **`run_autoformat.sh` reformats but does not check**, and CI's `autoformat` job runs
  that same script, so unformatted code does not fail CI — it is rewritten in the CI
  checkout and discarded. Run it locally and commit the result.
- **How selective CI actually selects.** `scripts/get_affected_packages.py` falls back
  to *all* packages whenever the change touches `.github/`, `scripts/`, a root
  `run_*.sh`, or any file outside a package directory. Otherwise it is the changed
  packages plus their transitive dependents — so a `kinder-models` change also runs
  `kinder-bilevel-planning`, `kinder-blockly`, `kinder-ds-policies` and
  `kinder-vlm-planning`.
- **Notebooks are not a CI job.** `kinder-bilevel-planning/run_ci_checks.sh` runs
  `--nbmake` locally and passes `--ignore=notebooks` to its pylint step; CI's `linting`
  job does not pass that flag, so notebooks are linted by CI and executed only locally.
- **Env-model discovery is by filename.** `env_models/__init__.py` uses
  `spec_from_file_location` over `kinematic2d/`, `dynamic2d/`, `kinematic3d/`,
  `dynamic3d/`; there is no registry to edit, and the caller's string must equal the
  module stem including case.
- **The test idiom**: end-to-end against the real simulator, one test function per
  file, fixed seeds, `MAKE_VIDEOS`/`RecordVideo`, the `for ... else: assert False`
  rollout, and abstract states asserted as one exact string of the whole sorted atom
  set — which is why there are no unit tests of individual predicates in the repo.
- **Three `bilevel_planning` behaviours that bite**: `TrajectorySamplingFailure`
  subclasses `BaseException`, so `except Exception` misses it; refinement succeeds only
  on `final_abstract_state == ns`, exact set equality of the whole abstract state, not
  "the add effects hold"; and `RelationalControllerGenerator` grounds a fresh
  controller — and so a fresh `PyBulletSim` — per sampling attempt.
- **Style rules that shape how code reads**, notably that `invalid-name` and the
  `[DESIGN]` limits are disabled, which is why CapWords locals like
  `OpenDrawerOperator` are idiomatic here, and that `duplicate-code` is off in
  `kinder-models` and has `min-similarity-lines=100` elsewhere, so a new env model that
  is a near-copy of its sibling is the accepted pattern.

## One thing to agree on before merging

`.gitignore` has ignored `CLAUDE.md` — unanchored, so at every depth — since the
initial import from prpl-mono (`b228c5c`). Tracking one means dropping that line, which
this PR does. If the intent was that each contributor keeps a private, untracked
`CLAUDE.md`, then the alternatives are to keep the ignore and put this content in
`CONTRIBUTING.md` instead, or to anchor the ignore to a different path and leave the
root file tracked. Happy to do either — say the word and I will re-cut it.

## Test plan

- [x] Docs-only plus a one-line `.gitignore` deletion; no source, config or test file
      is touched, so no package-level gate applies.
- [x] Every non-obvious claim in the file was verified against this tree rather than
      recalled. The two behavioural ones were checked by running them: pylint with and
      without `PYTHONPATH` (`E0013` + `10.00/10` versus the `C9900` the plugin is meant
      to raise), and the Projects-classic `NOT_FOUND` that breaks `gh pr edit` on this
      repo.
- [ ] CI: because the diff is entirely outside any package,
      `get_affected_packages.py` falls back to all ten packages, so this PR runs the
      full matrix.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -192,7 +192,6 @@ runs
 outputs
 output
 cleanrl
-CLAUDE.md
 
 # videos
 base_motion_planning.mp4
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1,0 +1,269 @@
+# CLAUDE.md
+
+Guidance for Claude Code (claude.ai/code) and for anyone else working in this repo.
+
+`kinder-baselines` holds baseline methods for the
+[KinDER](https://github.com/Princeton-Robot-Planning-and-Learning/kindergarden)
+benchmark. See `README.md` for the package list and the install steps; this file covers
+the conventions the code follows and the traps that are not visible from the config.
+
+## Layout
+
+Ten sibling packages, each self-contained: its own `pyproject.toml`, `.pylintrc`,
+`run_autoformat.sh`, usually `run_ci_checks.sh`, `src/<distribution_name>/`, and
+`tests/` mirroring `src/`. There is no root `pyproject.toml` and no shared settings
+file — every package carries its own copy of the whole toolchain config, including its
+own `pylint_plugins/` directory. A change to shared policy therefore has to be applied
+ten times.
+
+Packages depend on each other through `prpl_requirements.txt` (one relative path per
+line). `scripts/generate_topological_order.py` reads those to order installs;
+`scripts/install_all.py` uses that ordering, and `scripts/get_affected_packages.py`
+uses the reverse graph to decide what CI runs. Adding an intra-repo dependency means
+adding a line to `prpl_requirements.txt`, not just an import.
+
+Two packages are exceptions worth knowing before you write a script that assumes
+uniformity: `kinder-reward-grounding` has no `run_ci_checks.sh` (so
+`./run_all_ci_checks.sh` fails there), and `kinder-rl` and `kinder-reward-grounding`
+have no `pylint_plugins/` — `kinder-rl`'s `.pylintrc` has an empty `load-plugins=`.
+
+## Running the checks
+
+Run them from **inside the package directory**, not from the repo root:
+
+```bash
+cd <package-dir>
+./run_ci_checks.sh
+```
+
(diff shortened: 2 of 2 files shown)


joshnroy:
Closing: `CLAUDE.md` is in this repo's `.gitignore` (line 195, present since the initial prpl-mono extraction), and `kindergarden` ignores it too — so keeping these files untracked looks deliberate rather than accidental. Rather than remove the ignore, the conventions this documented are moving into our own downstream repo. Nothing here needs upstream action.

PR #98, opened 2026-08-07
kinder-pddlstream-planning: add the package and the Motion2D baseline

Minseo10:
Adds `kinder-pddlstream-planning`, a package of task and motion planning baselines built
on [PDDLStream](https://github.com/caelan/pddlstream), plus the first of three
environments. It carries the shared machinery the other two depend on.

Stacked on #97, which provisions the pddlstream checkout in CI. GitHub will retarget this
PR to `main` once that lands.

Six commits: scaffold, shared helpers, Motion2D domain/streams, Motion2D `run.py`,
Motion2D tests, docs.

**`ensure_pddlstream_importable()`** is the one piece of real design. It returns early if
`pddlstream` already imports, otherwise reads `PDDLSTREAM_PATH` (default `~/pddlstream`),
expands it, verifies the checkout really contains the package, and inserts it on
`sys.path`. A missing checkout raises a `ModuleNotFoundError` naming the path and the env
var rather than failing obscurely later. Each `run.py` calls it before importing anything
from `pddlstream`, which is why those files carry
`# pylint: disable=wrong-import-position,wrong-import-order` — the imports genuinely
cannot come first.

**`rendering.py`** holds `render_frame`/`save_gif`, shared by all three run scripts. GIFs
are written even when planning or execution fails, which is what makes a bad run
diagnosable.

**Motion2D** mirrors PDDLStream's `motion` example with one deliberate divergence worth
review: `connect` runs a full RRT-Connect query rather than the upstream straight-line
segment. A straight line is not a usable baseline in an environment built around narrow
passages.

Five tests cover action-bound conversion, open-field solving, GIF output, a single narrow
passage, and a multi-passage regression. Verified in the `kinder` env: mypy clean over 17
source files, pylint clean, `pytest tests/` 25 passed (the working tree also carries the
two environments still to come).

The scaffold commit is boilerplate copied from sibling packages. `pddlstream_utils.py` is
33 lines and is the part most worth scrutiny.


tomsilver (approved):
Really nice work! This opens up a lot of possibilities for us!

PR #99, opened 2026-08-07
kinder-pddlstream-planning: add the Packing3D baseline

Minseo10:
Stacked on #98. Adds the Packing3D baseline: pick parts up and seat them in a rack,
over `move_base`, `pick`, and `place`. The diff is scoped to `packing3d/`, its test file,
and a README section.

The streams supply the continuous parameters the symbolic layer leaves open — base poses,
grasps, and rack placements — checking each against IK and collisions before yielding, so
the planner never receives a geometrically impossible binding. The interesting constraint
is the rack: a placement must clear the walls on the way in while still ending up inside
the cavity, and those pull against each other as parts get larger.

`run.py` solves with the `adaptive` algorithm and takes `--num-parts`, `--seed`, and the
shared `--gif-path`:

    python -m kinder_pddlstream_planning.packing3d.run --num-parts 2 --seed 0

Four tests: `test_placement_candidates_clear_the_rack_walls` and
`test_placement_candidates_stay_inside_the_rack` (the two halves of that constraint, best
reviewed together), `test_grid_can_seat_every_part_at_once`, and
`test_placement_window_degrades_when_part_exceeds_cavity` — the negative case, where the
feasible window shrinks to nothing rather than silently yielding bad placements.

This is the first module to use `kinder_models`, already declared in the scaffold's
`prpl_requirements.txt`, so no dependency changes are needed. Rack geometry constants come
from `kinder_models.kinematic3d`, so placement numbers should be read against that module.

Verified in the `kinder` env: mypy clean over 17 source files, `pytest tests/` 25 passed.


tomsilver (approved):
LGTM! One thing to confirm before merging (by inspection) is that the new unit test doesn't add a huge amount of time to the overall unit tests

PR #100, opened 2026-08-07
kinder-pddlstream-planning: add the LimbRepositioning3D baseline

Minseo10:
Stacked on #99. The last of the three environments: grasp a human limb welded to the
gripper and torque it to a goal configuration, over `move_base`, `grasp`, and `move_limb`.
Sixteen variants — four scene types (isolated, human, wheelchair, bed) by four limbs.

Two things make this one harder than the other two.

**`move_limb` runs forward dynamics.** The action space is arm torques, so this cannot be
planned kinematically. `plan_limb_motion` runs a predictive-sampling MPC whose rollouts
are real PyBullet steps in the environment itself, not a surrogate. `check_torque_limits`
is a stream test rejecting trajectories past the robot's limits.

**The environment models no human collisions** — it zeroes the filter masks for the torso
and resting limbs, so the arm may be driven straight through the patient unless this
package supplies its own model. `LimbStreamContext` separates what the arm must avoid from
what the base must not be parked in, and treats the limb being repositioned as a target
rather than an obstacle. Since a seated person's parts already overlap at rest,
`refresh_collision_baseline` records the resting penetration per body and only counts
penetration beyond it — a check against zero would call the shipped start state a
collision.

**Per-variant overrides.** Four variants cannot be solved with the shipped placement: the
two `bed-*-leg` goals are only reachable from inside the bed (base-collision checking must
relax), and the two `wheelchair-*-arm` goals sit above the arm's vertical reach from any
floor position (world z must rise to 0.4 and 0.55, against the shipped 0.335 and 0.510).
Both are CLI flags, and `_VARIANT_OVERRIDES` in `run.py` applies them per variant so
`--all-variants` can sweep all sixteen in one invocation — without it the sweep passes one
setting to every variant and cannot reproduce the README's own instructions. Overrides
currently win over CLI flags during a sweep; single-variant runs are unaffected.

`--robot-base-z` translates the whole robot rigidly. It is a reachability experiment, not
a lift mechanism. Raising the shipped values upstream in `_ROBOT_PLACEMENTS` would affect
every consumer of the environment, so that is deliberately not done here — if those
placements are simply wrong, the right fix is upstream and this table should shrink.

    python -m kinder_pddlstream_planning.limbrepositioning3d.run --variant isolated-right-arm
    python -m kinder_pddlstream_planning.limbrepositioning3d.run \
        --all-variants --max-time 600 --gif-dir outputs/limbrepositioning3d

`solve_all_variants` records exceptions as failures so one bad variant cannot end the
sweep, and prints a summary table with per-variant wall time.

Sixteen tests cover placement and reachability (including that the shipped placement alone
cannot reach the goal, which is what makes `move_base` a real decision), grasping and state
restore, MPC progress and torque limits, the collision model above, and the override
resolution. Verified in the `kinder` env: mypy clean over 17 source files, pylint 17
passed / 25 deselected, `pytest tests/` 25 passed.

`stream.py` is the largest file in the package; the collision-model section and
`plan_limb_motion` are where the substance is.


Minseo10:
Closing for now — this is blocked on an upstream dependency, not on anything in this PR.

`limbrepositioning3d/stream.py` imports `NUM_LIMB_JOINTS` and `NUM_ROBOT_JOINTS` from
`kinder.envs.kinematic3d.utils`. Those constants were added in kindergarden commit
`f8ab168` ("Add limb object types and torque-control utilities"), which currently lives
on `origin/limb/02-limb-env` and `origin/limb/03-limb-docs` but has **not** reached
kindergarden's `main`. Since this package declares `kindergarden>=0.1.0`, CI resolves a
published version that predates the symbols, and both mypy and the tests fail:

    stream.py:19: error: Module "kinder.envs.kinematic3d.utils"
                  has no attribute "NUM_LIMB_JOINTS"  [attr-defined]

The code passes locally against a kindergarden checkout on the `limb/*` branches.

I will reopen this once the limb work lands on kindergarden `main` and is released. The
branch `pddlstream/03-limbrepositioning3d` is kept, so nothing is lost.

Motion2D and Packing3D are unaffected and continue in #98 and #99.

PR #101, opened 2026-08-07
README: list kinder-pddlstream-planning in the package table

Minseo10:
Stacked on #99, and intentionally last in the stack.

Adds one row to the root package table:

    | `kinder-pddlstream-planning` | PDDLStream task and motion planning baselines |

This line was originally part of #98, but the root `README.md` maps to no package, so
`scripts/get_affected_packages.py` hits its "files changed outside any package" safety
fallback and marks all eleven packages affected. One documentation line was pulling the
entire monorepo suite into #98 and exposing it to unrelated flaky tests.

Splitting it out lets #98 and #99 run only `kinder-pddlstream-planning`. This PR still
triggers the full run, which is fine — it is a single-line documentation change, so a
flaky failure elsewhere just needs a re-run.

Landing last also means the table never advertises a package that is not yet on `main`.


tomsilver (approved):
Woohoo!

Issue #105, opened 2026-08-12
Remove stale Motion3D and Ground3D (kinematic3d) models, skills, and configs


tomsilver:
kindergarden removed the `Motion3D` and `Ground3D` **kinematic3d** environments in Princeton-Robot-Planning-and-Learning/kindergarden#120 (July 2026), deleting `kinder/envs/kinematic3d/motion3d.py` and `ground3d.py` along with their registrations and docs.

kinder-baselines still carries models, skills, tests, and configs for both. They import cleanly today only because the environments are pinned to an older kindergarden — `kinder-bilevel-planning/.venv` currently has `kindergarden==0.1.3`, which still ships those modules. On any current kindergarden they fail at import.

Found while adding models for the new `VegaMotion3D` environment.

## Files to remove

**kinder-bilevel-planning**
- `src/kinder_bilevel_planning/env_models/kinematic3d/motion3d.py`
- `src/kinder_bilevel_planning/env_models/kinematic3d/ground3d.py`
- `tests/env_models/geom3d/test_motion3d.py`
- `tests/env_models/geom3d/test_ground3d.py`
- `experiments/conf/env/motion3d.yaml`

**kinder-models**
- `src/kinder_models/kinematic3d/motion3d/parameterized_skills.py`
- `src/kinder_models/kinematic3d/ground3d/parameterized_skills.py`
- `tests/kinematic3d/motion3d/test_geom3d_parameterized_skills.py`
- `tests/kinematic3d/ground3d/test_ground3d_parameterized_skills.py`

## Files needing edits rather than deletion

- `kinder-models/scripts/planning_data_kinematic3d.py` — imports `ObjectCentricGround3DEnv` and `ObjectCentricMotion3DEnv` at lines 10-11 and branches on them around line 64. Drop those branches; the `BaseMotion3D` path stays.
- `kinder-rl/src/kinder_rl/dense_rewards.py:133` — a `"kinder/Motion3D"` key mapping to `BaseMotion3DDenseReward`. Dead lookup, since no such env id registers any more.

## Explicitly *not* stale

`kinder-models/src/kinder_models/dynamic3d/arm_controller.py` and `base_controller.py` import `kinder.envs.dynamic3d.motion3d`, a **different** module that still exists. Leave them alone. Likewise anything matching `BaseMotion3D`, which is a live environment.

## Suggested follow-up

[...]

PR #106, opened 2026-08-12
Add bilevel planning models for VegaMotion3D

tomsilver:
Bilevel planning models for the new `VegaMotion3D` environment (Princeton-Robot-Planning-and-Learning/kindergarden#140), laid out like the TidyBot models: skills and abstractions in `kinder-models`, `SesameModels` wiring in `kinder-bilevel-planning`.

The environment has exactly one skill — move the arm to the target — so the models are pure motion planning: sample an end-effector pose at the target, solve IK for a goal configuration, plan a collision-free joint path, emit it as bounded joint deltas.

## OMPL by default

Motion planning goes through `prpl_kinematics` rather than `pybullet_helpers`, since that is what backs the environment. `create_motion_planner` returns an `OMPLPlanner` when `ompl` is importable and a `BiRRTPlanner` otherwise; both satisfy the same `MotionPlanner` protocol. `prefer_ompl=False` forces the fallback, so the tests are parameterized over both and the BiRRT path stays covered even where `ompl` is installed.

`ompl` is an extra of `prpl_kinematics` (`prpl_kinematics[planning]`) precisely because it ships wheels for fewer platforms, so "use it if installed" is the right default rather than a hard requirement.

Both solve the environment locally:

```
prefer_ompl=True  -> solved in 43 steps, 0.43s
prefer_ompl=False -> solved in 34 steps, 0.33s
```

## Why the orientation sampler is a cone, not uniform

The older `Motion3D` skills sampled end-effector orientation uniformly over SO(3). Copying that here would mostly produce sampling failures. Vega's analytic IK solves a non-SRS 7R arm, so orientation largely decides whether a reachable-looking target admits a solution at all. Measured over the environment's target bounds:

| Orientation sampling | IK success |
|---|---|
| uniform over SO(3) | 27% |
| within 30° of home (**chosen**) | 95% |
| home orientation only | 97% |

A cone keeps the sampler diverse without spending most draws on failures. The width is `DEFAULT_ORIENTATION_PERTURBATION` and is per-controller configurable.

## Abstractions

`AtTarget` defers to the environment's own `goal_reached()` instead of recomputing the distance threshold. Whether the end effector is on target depends on forward kinematics rather than state features alone, so recomputing it here would duplicate the threshold and invite drift from the environment. This is why the abstractor is a factory closing over a sim rather than a bare function as in the TidyBot models.

## Packaging

`prpl_kinematics` is an optional extra in both packages, mirroring how kindergarden treats it, and is in `develop` so CI covers the new code. `env_models/__init__.py` gains `kinematic3d_v2` in its lookup path.

## Verification

New tests pass locally against kindergarden `main` plus `prpl_kinematics[planning]`: 4 in kinder-models, 2 in kinder-bilevel-planning, covering the controller end to end on both planners, that sampled parameters really put the end effector in range, and planner selection.

**The tests skip in CI for now.** `kindergarden==0.2.0`, the current release, predates `kinematic3d_v2`, so `pytest.importorskip` skips the modules. They activate on the next kindergarden release with no change here.

## Note

That same release will break the stale `Motion3D`/`Ground3D` models — see #105, which I updated with the details while working on this. Pointing a venv at kindergarden `main` today produces 8 mypy errors and 7 test failures across these two packages, all confined to those files and none from this PR.

### DIFF
--- a/kinder-bilevel-planning/pyproject.toml
+++ b/kinder-bilevel-planning/pyproject.toml
@@ -25,7 +25,12 @@ dependencies = [
 ]
 
 [project.optional-dependencies]
+# Backs the kinematic3d_v2 env models. See the same extra in kinder-models.
+prpl-kinematics = ["prpl_kinematics[planning]>=0.0.1"]
 develop = [
+    # Installed for development and CI so the kinematic3d_v2 models are linted, type
+    # checked, and tested.
+    "prpl_kinematics[planning]>=0.0.1",
     "black",
     "docformatter",
     "ipykernel",
@@ -71,6 +76,10 @@ exclude = ["venv/*", ".venv/*", "third-party/*"]
 [[tool.mypy.overrides]]
 module = [
     "matplotlib.*",
+    # Optional; see the prpl-kinematics extra.
+    "prpl_kinematics.*",
+    "spatialmath.*",
+    "pybullet.*",
     "mujoco",
     "mujoco.*",
     "ruckig",
--- a/kinder-bilevel-planning/src/kinder_bilevel_planning/env_models/__init__.py
+++ b/kinder-bilevel-planning/src/kinder_bilevel_planning/env_models/__init__.py
@@ -21,6 +21,7 @@ def create_bilevel_planning_models(
         current_file.parent / "kinematic2d" / f"{env_name}.py",
         current_file.parent / "dynamic2d" / f"{env_name}.py",
         current_file.parent / "kinematic3d" / f"{env_name}.py",
+        current_file.parent / "kinematic3d_v2" / f"{env_name}.py",
         current_file.parent / "dynamic3d" / f"{env_name}.py",
     ]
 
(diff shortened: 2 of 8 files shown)


tomsilver:
Updated now that kindergarden **0.2.1** is published with `kinematic3d_v2`.

**The tests no longer skip.** The PR body's claim that they would skip in CI was true against 0.2.0 and is now obsolete. Against 0.2.1 they run and pass locally:

```
kinder-bilevel-planning/tests/env_models/kinematic3d_v2   2 passed
kinder-models/tests/kinematic3d_v2                        4 passed
```

The `pytest.importorskip` guards stay, since `prpl_kinematics` is still an optional extra of both packages; only the "not in a release yet" half of the reasoning went away, and the comments are corrected to say so.

**This PR needs #107 to land first.** 0.2.1 also removes the `Motion3D`/`Ground3D` kinematic3d environments, which breaks the stale models here (8 mypy errors, 7 test failures) — unrelated to this change but enough to keep CI red. #107 covers it, including the `kindergarden>=0.2.1` floor bumps, so I have deliberately left dependency floors untouched here to avoid conflicting with it.

Earlier CI failures on this PR were a separate, now-stale cause: that run resolved kindergarden 0.2.0, so mypy and pylint could not resolve `kinder.envs.kinematic3d_v2` at all. `importorskip` guards the tests but not static analysis. A re-run after #107 merges should be clean.

PR #113, opened 2026-08-13
Tossing3D to two skills and five predicates + Bilevel planning

joshnroy:
# Open Questions

- [ x] Why does the arm spin in the plan? it's in every seed

^^ This is now fixed

# Results

https://github.com/user-attachments/assets/9f6d7286-b2ed-4263-b4ff-4728b3adbcef

Succeeds 100/100 times

# For Reviewer

This should be fairly in depth and may honestly be easier to walk through together. I make a list of things I _think_ are happening here, but I may not be following the best conventions and not all of the code may be needed

This goes with PR https://github.com/Princeton-Robot-Planning-and-Learning/kindergarden/pull/162

# Predicates

1. `MovableInGoalRegion` -> Cube is a `Movable` and this checks if it is successfully in the "bin" which is represented by a goal region
2. `OnGround` -> this checks if the cube is "on the ground" aka - it is flat (see note on cube symmetry later) and within sufficient distance of the floor
3. `Holding` -> Simple. The robot is holding the cube
4. `HandEmpty` -> The robot is not holding the cube
5. `MovableIsDownX` -> In reality, this is used to see if the Cube (`Movable`) is on the robot side (Down `X` axis of) the Barrier (second argument `Movable`)

# Skills

## Pick Up Cube

Moves to the cube and picks it up, has continuous parameters of distance and rotation for the grasp, following the shelf example. Theoretically, we should be able to calculate these, but it proved to be sufficiently time consuming to hardcode them robustly and easy to let the bilevel planner plan them. Plus, it is a "harder" benchmark if they are not hardcoded

### Phases

1. Open the gripper (if it's closed, otherwise no-op): this will be important once we get to reset-free practice
2. Move base to the fixed-offset grasping location
3. Motion plan the hand to a position above the cube, with the correct yaw to be aligned with the cube
4. Move the hand straight down so that the grippers are around the cube
5. Close the grippers
6. Bring the cube back to home state

We separate the grasping steps because it was otherwise difficult to avoid the robot accidentally hitting the cube on its way around it. Though technically not necessary, forcing the end effector to first align with the cube proved easier

### Cube Orientation (also applies to `OnGround`)

We don't actually want to yaw the end effector relative to the cube's canonical orientation. Ex: if it is upside down, the gripper should not try to approach it from there floor. Therefore, we define an operator that allows the end effector to pretend the cube is "upright" at all times.

## Toss Cube

Moves to the tossing position and throws the cube. Technically, this could have been two skills, but defining the correct "Tossing Region" is not workable without back-calculating from the throw. 4 parameters is simple enough that we can combine them.

### The toss itself

The toss is parameterized by tossing speed (degrees per second) and the number of ms into the toss. This is the same parameter space we have access to on the real kinova arm.

1. Motion plan the trajectory between the starting pose (hard coded) and end pose (hard coded)
2. Replace the waypoints with a trapezoidal velocity controller. a. constant acceleration. b. constant velocity (during which the toss should happen). c. constant deceleration


yixuanhuang98 (approved):

kinder-models/src/kinder_models/dynamic3d/tossing/parameterized_skills.py
```diff
@@ -114,14 +127,14 @@ def reset(
 
     def terminated(self) -> bool:
         assert self._current_base_motion_plan is not None
-        return self._robot_is_close_to_pose(self._current_base_motion_plan[-1])
```
joshnroy: These are just renames of the same function

kinder-models/src/kinder_models/dynamic3d/tossing/parameterized_skills.py
```diff
@@ -318,20 +331,28 @@ def __init__(self, *args, **kwargs) -> None:
         self._current_params: np.ndarray | None = None
         self._current_arm_joint_plan: list[JointPositions] | None = None
         self._pybullet_sim: PyBulletSim | None = None
-        # Fraction of toss path at which to release gripper.
```
joshnroy: Part 1 of abstraction. I changed two things about the swing:
1. it now releases at a specific time (parameter) and toss speed (parameter) rather than a (hardcoded) percentage of the swing and speed. This is to match the arguments of the kinova controller.
2. The actual swing logic is abstracted to a helper

kinder-models/src/kinder_models/dynamic3d/tossing/parameterized_skills.py
```diff
@@ -353,67 +374,37 @@ def reset(self, x: ObjectCentricState, params: Any) -> None:
         )
         assert plan is not None, "Motion planning failed"
         self._current_arm_joint_plan = plan
-        # Compute trapezoidal velocity profile along the path.
```
joshnroy: Part 2 of abstraction. I changed two things about the swing:
1. it now releases at a specific time (parameter) and toss speed (parameter) rather than a (hardcoded) percentage of the swing and speed. This is to match the arguments of the kinova controller.
2. The actual swing logic is abstracted to a helper

kinder-models/src/kinder_models/dynamic3d/tossing/parameterized_skills.py
```diff
@@ -353,67 +374,37 @@ def reset(self, x: ObjectCentricState, params: Any) -> None:
...
         self._step_idx = 0
 
     def terminated(self) -> bool:
         # Terminate when we've gone through the entire profile.
-        return self._step_idx >= len(self._trajectory)
+        assert self._swing is not None
+        return self._step_idx >= len(self._swing.trajectory)
 
     def step(self) -> Array:
-        assert self._current_arm_joint_plan is not None
```
joshnroy: Part 3 of abstraction. I changed two things about the swing:
1. it now releases at a specific time (parameter) and toss speed (parameter) rather than a (hardcoded) percentage of the swing and speed. This is to match the arguments of the kinova controller.
2. The actual swing logic is abstracted to a helper

kinder-models/src/kinder_models/dynamic3d/tossing/parameterized_skills.py
```diff
@@ -697,19 +690,390 @@ def _get_current_gripper_pose(self) -> float:
...
 
-    def _robot_gripper_is_open(
+    def _check_robot_gripper_is_open(
         self, atol: float = GRIPPER_OPEN_COMMAND_TOLERANCE
     ) -> bool:
         current_gripper_pose = self._get_current_gripper_pose()
         return current_gripper_pose < atol
 
 
+class PickCubeControllerPhase(enum.Enum):
```
joshnroy: I started using enums to keep track of the "Phases" of each multi stage controller rather than a set of booleans

kinder-models/src/kinder_models/dynamic3d/tossing/parameterized_skills.py
```diff
@@ -697,19 +690,409 @@ def _get_current_gripper_pose(self) -> float:
...
+        cube: The cube to pick up.
+
+    Where to stand is derived rather than sampled: STANDOFF_LADDER is walked from the
+    nominal pose outwards until one plans, so a caller cannot draw an unreachable pose
+    and there is nothing for a refiner to backtrack over. A grasp that closes on
+    nothing releases before terminating, leaving the hand empty rather than commanded
+    shut on air.
+    """
+
+    class PickCubeControllerPhase(enum.Enum):
```
joshnroy: I started using enums to keep track of the "Phases" of each multi stage controller rather than a set of booleans

kinder-models/src/kinder_models/dynamic3d/cube_symmetry.py
```diff
@@ -0,0 +1,65 @@
+"""The rotations that map a cube onto itself, and what follows from them.
```
yixuanhuang98: Could we move this to utils.py?

yixuanhuang98 (approved):
Overall, LGTM! Left one comment.

PR #119, opened 2026-08-14
kinder-models: extract the simulator-free predicate checks into their own module

joshnroy:
## Summary

The dynamic3d state abstractors answer most of their predicates with a few lines of arithmetic over pose features, but there is no way to ask for one of them. The only entry point is `Tossing3DStateAbstractor(sim).state_abstractor(state)`: a stateful class whose `__init__` calls `sim.reset()` and constructs a `PyBulletSim`, which answers for every predicate at once, and which needs an `ObjectCentricState` a consumer with its own state representation cannot construct. So consumers reimplement the arithmetic instead, and then it drifts.

This moves that arithmetic, and the five tolerances it reads, into `kinder_models/dynamic3d/predicate_checks.py`. Two properties, both deliberate:

- **It takes plain floats**, not an `ObjectCentricState`, so a caller holding some other state representation can use it.
- **It imports numpy and nothing else**, so importing it pulls in neither `kinder.envs`, PyBullet nor MuJoCo. `test_importing_this_module_pulls_in_no_simulator` asserts that in a subprocess — necessarily, since the test file's own other imports would otherwise have loaded the simulator into the interpreter already.

The tolerances **move** rather than being re-exported from `utils.py`. `utils.py` exports `PyBulletSim`, so importing a threshold from there costs a PyBullet import; re-exporting would have kept that cost while hiding it. Every importer moves with them (`shelf`, `sweep3D` and `tossing` skills and abstractors) — five files, import lines only.

Behaviour is unchanged. The Tossing3D abstractor now calls in here for four of its six checks and keeps the two that genuinely need the simulator: the forward kinematics behind `Holding`, and the goal-region lookup behind `MovableInGoalRegion`. `_check_holding` is split accordingly — `check_grasped_and_lifted` is the simulator-free pair of conjuncts, `check_end_effector_at_object` is the comparison against an end-effector pose the caller supplies.

Not done here, and worth a separate look: `shelf/state_abstractions.py` still writes the same on-ground and grasped-and-lifted arithmetic out by hand. It is behaviour-identical to what moved, but refactoring it touches Shelf3D, so it is left out of a PR that is otherwise additive.

## Test plan

- [x] `pytest tests/dynamic3d/test_predicate_checks.py -q` — 8 passed (new file)
- [x] `pytest tests/dynamic3d/tossing/test_tossing_state_abstractions.py -q` — 11 passed, unchanged from before this PR
- [x] `pytest tests/dynamic3d/shelf tests/dynamic3d/sweep3D tests/dynamic3d/test_dynamic3d_utils.py -q` — 13 passed, covering the five files whose imports moved
- [x] `pylint --rcfile=.pylintrc <changed files>` — 10.00/10, with `PYTHONPATH` set to the package directory so `pylint_plugins.no_np_random` actually loads (verified by probing a file containing `np.random.rand()` and seeing `C9900` fire; without the plugin pylint reports `E0013` and still scores 10.00)
- [x] `mypy . --no-incremental` — no new errors; the 3 reported are pre-existing `import-not-found` for `astroid`/`pylint` stubs in `pylint_plugins/`, absent from the environment used
- [x] `black`, `isort`, `docformatter` via the package's own `run_autoformat.sh` tooling, on the changed files only

## Followup

- `shelf/state_abstractions.py` and `sweep3D/state_abstractions.py` can call the same helpers.
- CI on this branch currently fails at collection, inherited from the open stack below it being blocked on a `kindergarden` PyPI release. Expected, and unrelated to this change.

### DIFF
--- a/kinder-models/src/kinder_models/dynamic3d/predicate_checks.py
+++ b/kinder-models/src/kinder_models/dynamic3d/predicate_checks.py
@@ -1,0 +1,86 @@
+"""The numeric cores of the dynamic3d state abstractions, free of any simulator import.
+
+Every dynamic3d state abstractor answers most of its predicates with a few lines of
+arithmetic over pose features, and those lines are the part worth sharing: shelf's
+abstractor and tossing's carry the same gripper-open test, the same on-ground test and
+the same grasped-and-lifted test, written out twice.
+
+Two properties, both deliberate, and both the reason this is a module rather than more
+private methods on the abstractors:
+
+- **It takes plain floats, not an `ObjectCentricState`.** An abstractor is the only way
+  to ask upstream whether a predicate holds, it answers for every predicate at once, and
+  it needs a state type a consumer with its own state representation cannot construct.
+  Arithmetic over floats has none of those constraints.
+- **It imports numpy and nothing else**, so importing it pulls in neither `kinder.envs`,
+  PyBullet nor MuJoCo. `utils.py` cannot offer that -- it exports `PyBulletSim` -- which
+  is why the tolerances moved here and are re-exported from there rather than the other
+  way around. `test_predicate_checks.py` asserts the property in a subprocess, so it
+  fails loudly if a future import erodes it.
+
+What is *not* here is anything needing the simulator: the forward kinematics behind
+`Holding` and the goal-region lookup behind `MovableInGoalRegion` both do, so the
+abstractor keeps them and calls in here for the rest.
+"""
+
+from typing import Sequence
+
+import numpy as np
+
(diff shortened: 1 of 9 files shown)


joshnroy:
Closing as out of scope for now, at Josh's direction: this is a cross-cutting refactor of `shelf/`, `sweep3D/`, `tossing/` and `utils.py` (9 files), and the constraint that motivated it — a downstream consumer needing predicate arithmetic it could import without pulling in `kinder.envs` — has since been lifted, so it buys something already paid for.

Reopenable if the module-scope import route proves unworkable downstream. Branch `josh/feature/dynamic3d-pure-predicate-checks` left in place; nothing is deleted.

PR #120, opened 2026-08-14
kinder-models: derive RobotAtThrowPose from the scored region, not the sampler's range

joshnroy:
## Summary

`RobotAtThrowPose` accepted any standoff inside `THROW_STANDOFF_BOUNDS = (1.20, 1.65)`, which is the range a base sampler draws from. On the bilevel-planning branch stacked above this one, `TOSS_TARGET_DISTANCE_BOUNDS = (1.25, 1.45)` sits strictly inside it, and its own comment says as much:

```python
# Not the (0.5, 0.6) grasping range. Sits inside THROW_STANDOFF_BOUNDS, so every draw
# satisfies RobotAtThrowPose.
```

`MoveToThrowPose`'s only add effect is `RobotAtThrowPose`. So that containment makes the add effect hold after every attempt by construction: the label is constant-true, a per-skill success classifier trained on it sees one class, and the sampler never improves on its uniform prior — with nothing about the run looking wrong. This is not hypothetical. A downstream port of this domain (`hitl-pmp`) shipped the same predicate, measured `16/16` attempts labelled success against `0/16` informed draws — versus `7/20` informed for a pick skill in the same run — and spent four PRs removing it.

**The fix is to ask a different question**: not "is this standoff one a sampler could produce", but "does the throw's own displacement carry the object into the scored region". The region is read live off `blocks_goal_region`, so the accepted band follows the bin and the region when either moves, rather than being a literal that goes stale. The interval form (`check_reach_interval_hits_box`) takes a reach *interval* so a controller with tunable toss parameters can ask whether *some* parameterisation scores; the shipped toss is fixed, so it passes the degenerate `reach_min == reach_max`.

`THROW_STANDOFF_BOUNDS` stays, re-commented, as the sampler's range and explicitly *not* the predicate's acceptance interval — the two must not be re-coupled or the tautology returns silently.

**One constant is calibrated rather than derived**: `THROW_REACH = 1.275`, the distance the shipped toss carries the cube before first ground contact. It is a property of the controller (the two fixed arm configurations, the cube's mass), not of the scene. It comes from downstream calibration against *where success breaks*, not from free flight — throwing onto open floor and recording where the cube rests gives 1.3499 m (sd 0.0024, n = 12), ~0.075 m longer because it includes post-impact roll, and the region test needs the impact range. That calibration was taken with the gripper opening on the first control step past release fraction 0.46, which is exactly what `TossController` does here.

**The downstream reliability margins are deliberately not copied.** `hitl-pmp` narrows the box by 0.025 m and 0.05 m, because a 5-seed sweep found the standoffs solving on every seed narrower than the geometric prediction at both ends. Those margins are indexed by *commanded* standoff, and commanded standoff is not what this predicate reads: driving `move_to_target` to a commanded 1.35 m leaves the base 1.3778 m from the bin — 27.8 mm short, comfortably inside `WAYPOINT_TOLERANCE`, but a third of what either margin trims. Transplanting them would import a 28 mm indexing error along with the measurement. Measured on this branch, `seed=125`:

| commanded standoff | achieved base x | achieved standoff | predicted landing x |
| --- | --- | --- | --- |
| 1.35 | 0.6222 | 1.3778 | 1.8972 |
| 1.60 | 0.3727 | 1.6273 | 1.6477 |

**Note for the branch stacked above.** With this landed, `TOSS_TARGET_DISTANCE_BOUNDS = (1.25, 1.45)` is no longer constant-true — draws above ~1.375 are rejected — which is the point, but it does change what that sampler's label distribution looks like. Nothing on that branch is touched here.

## Test plan

All run against a real simulator, `seed=125`.

- [x] `pytest tests/dynamic3d/tossing/test_tossing_state_abstractions.py -q` — 15 passed
- [x] `pytest tests/dynamic3d/tossing tests/dynamic3d/test_predicate_checks.py -q` — 38 passed
- [x] New `test_robot_at_throw_pose_fails_at_a_standoff_the_sampler_can_draw`: drives to 1.60 m, inside `THROW_STANDOFF_BOUNDS`, and asserts the atom is **absent**. Fails against the pre-fix predicate — which is the defect stated as a test.
- [x] New `test_robot_at_throw_pose_agrees_with_whether_the_throw_scores`: full pick → move-to-standoff → windup → toss at 1.35 and 1.60, asserting the predicate and `_check_goals()` give the same verdict at each. This is the calibration guard for `THROW_REACH`, against real episode outcomes rather than a recorded number.
- [x] New `test_the_accepted_band_follows_the_goal_region_rather_than_a_literal`: shifting the region 1 m shifts the accepted base positions by the same metre.
- [x] Existing `test_robot_at_throw_pose_holds_after_driving_to_a_throw_standoff` (1.35 m) still passes, unmodified.
- [x] Both new behavioural tests confirmed **red** against the pre-fix check body before the fix, with the rest of the diff in place.
- [x] `pylint --rcfile=.pylintrc <changed files>` — 10.00/10, plugin load proven as in the PR below.
- [x] `mypy . --no-incremental` — no new errors.

## Followup

- The heading conjunct is unchanged and still uses `THROW_POSE_TOLERANCE`; only the standoff conjunct moved.
- A consumer with its own reliability measurement can pass an already-trimmed box rather than the region's raw edges.

### DIFF
--- a/kinder-models/src/kinder_models/dynamic3d/predicate_checks.py
+++ b/kinder-models/src/kinder_models/dynamic3d/predicate_checks.py
@@ -84,3 +84,30 @@ def check_end_effector_at_object(
 def check_is_down_x(x: float, other_x: float) -> bool:
     """Whether a movable is at lower x than another, read live rather than fixed."""
     return bool(x < other_x)
+
+
+def check_reach_interval_hits_box(
+    base_x: float,
+    reach_min: float,
+    reach_max: float,
+    box_x_min: float,
+    box_x_max: float,
+) -> bool:
+    """Whether a throw from base_x can land inside [box_x_min, box_x_max].
+
+    A throw from a fixed windup displaces the object by a distance set by the toss
+    parameters and not by the scene, so the set of x it can reach from base_x is the
+    interval [base_x + reach_min, base_x + reach_max]. The pose is throwable-from iff
+    that interval meets the box -- iff *some* toss parameterisation scores from here.
+    A controller with a single fixed toss passes reach_min == reach_max, which
+    degenerates to a point and is the same test.
+
+    **Deriving the acceptance band this way, rather than accepting the range the base
+    sampler draws from, is the whole point of the function.** The two are easy to
+    conflate and the failure is silent: if a "robot is at a throw pose" predicate
+    accepts exactly the standoffs the sampler can produce, then the move-to-throw-pose
+    operator's only add effect holds after every attempt by construction. A learner
+    that trains a per-skill success classifier on that label sees a single class and
+    never improves on its uniform prior, and nothing about the run looks wrong.
+    """
+    return bool(base_x + reach_min <= box_x_max and box_x_min <= base_x + reach_max)
--- a/kinder-models/src/kinder_models/dynamic3d/tossing/state_abstractions.py
+++ b/kinder-models/src/kinder_models/dynamic3d/tossing/state_abstractions.py
@@ -33,6 +33,7 @@ from kinder_models.dynamic3d.predicate_checks import (
     check_gripper_open,
     check_is_down_x,
     check_on_ground,
+    check_reach_interval_hits_box,
 )
 from kinder_models.dynamic3d.utils import (
     WAYPOINT_TOLERANCE,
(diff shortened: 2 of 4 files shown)


joshnroy:
Superseded: the `RobotAtThrowPose` derivation now lives in #113 itself, as `540ced2`, which is where the throw-pose controllers it describes already live. The version here additionally depended on #119's extraction; the version in #113 is self-contained, so it adds no new dependency to a PR on the release critical path.

The constant also changed as a result of being re-measured on #113's own line rather than carried over from the downstream port: `THROW_REACH` is 1.2285 here, not 1.275. Branch left in place.

PR #135, opened 2026-08-19
Add optional top-down grasp constraint to VegaPickPlace3D pick and place

### LINKED ISSUE #134: Grasp sampling is orientation-blind; add optional top-down grasp constraint
## Problem

`_sample_ee_reach_configuration` (kinder_models/kinematic3d_v2/vega_pickplace3d/parameterized_skills.py) draws uniformly random joint configurations and keeps the draw's orientation through the differential-IK refinement. Grasp orientation is therefore whatever random sampling happens to produce. On the real Vega this yields grasps at arbitrary tilts; for tabletop pick-and-place we want the option of top-down grasps (fingers pointing down, yaw free).

Rejection alone cannot fix this: a uniform draw points the end effector within 15 deg of vertical roughly 2% of the time, before the reach condition. In a 250k-draw experiment at the real-table grasp pose (cube at (0.46, -0.16, 0.647) base frame, table top z 0.617), zero draws satisfied tilt <= 12 deg plus reach.

Top-down grasps do exist there: a direct IK sweep (down-facing target, yaw swept, `NumericalIK` over the same manipulator/ee frame the sampler uses) found 468 configurations with 0.0 deg tilt that are self-collision-free, table-clear, and keep the wrist roll well inside its soft limits. They all require the shoulder-forward posture (right j1 near 0 rather than home's -1.81), which uniform sampling near typical seeds essentially never reaches together with the orientation condition.

## Proposal

Add an optional grasp-orientation constraint, e.g. `grasp_approach_max_tilt` (radians off vertical; None = current behavior), plumbed through `create_bilevel_planning_models` so consumers can set it per pipeline.

When set, `GroundPickController.sample_parameters` should sample constraint-aware instead of filtering: sample a yaw, build the down-facing target pose over the cube, solve full IK for it (the sampler already owns a `NumericalIK` per side), then apply the existing collision check. Fall back to the current uniform draw when the constraint is unset. `GroundPlaceController` wants the same treatment so set-downs are also fingers-down.

Happy to send a PR for this — the IK-sweep prototype (which found the 468 configurations) can serve as the sampler implementation directly.

Context: first hardware pick-and-place runs on the PRPL Vega (prpl-dexmate); the operator rejected the sampled grasp at the confirm gate because of its tilt.

tomsilver:
Closes #134.

## What this does

Adds an optional `grasp_approach_max_tilt` constraint (radians off vertical; `None` = current behavior) to the VegaPickPlace3D skills. When set, `GroundPickController` and `GroundPlaceController` sample goal configurations constraint-aware instead of by rejection:

1. Draw a yaw uniformly and build the down-facing end-effector target rotation (the Vega tool frames carry the fingertip offset along their z axis, so "fingers down" is frame z along world -z).
2. Solve full-pose IK for that target with `NumericalIK`, seeded from a uniform joint draw. The solver's orientation tolerance is the tilt bound itself, so a converged solve satisfies the constraint by construction.
3. Apply the existing acceptance checks: strict grasp-range reach for pick, the release slab over the target patch for place, and the collision check.

For place, the end-effector target position accounts for the held cube's fixed offset in the end-effector frame, rotated by the candidate target rotation, so the cube (not the gripper) lands over the patch.

The parameter is plumbed through `create_lifted_controllers` and `create_bilevel_planning_models` (the `env_models` dispatcher already forwards kwargs, so top-level callers can pass it directly). Handover sampling is unconstrained: the take-over happens in mid-air, where approach direction does not matter.

## Why constraint-aware instead of rejection

As measured in #134, a uniform draw is within 15 deg of vertical about 2% of the time before the reach condition, and zero of 250k draws satisfied tilt <= 12 deg plus reach at the real-table grasp pose. Rejection cannot be salvaged by budget.

## Budget and iteration choices (measured on this sampler)

Each constrained candidate is a full IK solve (~0.2 s) rather than an FK call (~0.3 ms), so the constrained budget is separate: `DOWN_FACING_NUM_GOAL_CANDIDATES = 100`.

- Per-candidate acceptance for reachable cubes (tilt bound 0.1 rad, `max_iters=100`, soft-limit margin applied): 46% for seed 0 / right arm, 10% for the harder seed 19 / left arm. At 10%, 100 candidates leave a spurious-exhaustion probability under 3e-5.
- A genuine exhaustion (cube out of the arm's reach) resolves in ~25 s at `max_iters=100`. Bilevel planning relies on that failure to discard wrong-arm abstract plans; a spurious failure is the costlier mistake, so the budget favors reliability.
- `max_iters=50` would halve the failure-path cost but drops hard-case acceptance from 20% to 16%; `DOWN_FACING_IK_MAX_ITERS = 100` keeps the better acceptance.

## Tests

- `test_top_down_pick_and_place_controllers`: constrained pick and place reach the environment goal, with both sampled configurations verified within the tilt bound and the grasp within grasp radius.
- `test_top_down_pick_sampling_fails_for_out_of_reach_arm`: the constrained sampler's (separate) exhaustion path still raises `TrajectorySamplingFailure`.
- `test_vega_pickplace3d_grasp_approach_max_tilt_plumbing` (kinder-bilevel-planning): the kwarg threads through model creation to the pick sampler, verified by tilt of the sampled configuration.

CI: mypy, pylint, and black are clean in both packages. Full kinder-bilevel-planning suite passes (80 tests + notebook). Full kinder-models suite: all vega_pickplace3d tests pass; 22 failures in unrelated environments (tossing, obstruction3d, shelf3d, transport3d, test_utils) reproduce identically on a clean checkout of main in my environment (e.g. TidyBot task-config paths resolving into site-packages), so they are pre-existing and not caused by this change.

## Not addressed

- The constrained sampler keeps `NumericalIK` with random restarts, as proposed in the issue. I benchmarked Vega's attached analytic solver (`VegaArmIK`, via `sim.robot.manipulators[side].ik`) as an alternative: it solves exactly-reachable poses reliably (5/5 on FK-derived controls, ~40 ms per query) but returned 0 solutions in 150 attempts on the sampler's actual targets, because it is an exact solver and the down-facing targets are only reachable within the sampler's tolerances (4 cm position, the tilt bound). The tolerance slack is what makes top-down grasps feasible across the table, and the numerical solver exploits it natively; using the analytic solver would require discretizing that slack into many exact queries, a different design with no demonstrated win.
- Handover take-over orientation is left unconstrained (see above).
- `grasp_approach_max_tilt` must be positive when set: it doubles as the IK orientation tolerance, so exactly zero is unsatisfiable. An assert guards this.
- The repo's committed formatting disagrees with the current venv formatter versions (black 26.x / isort 8.x / docformatter 1.7.8) on ~14 files untouched here; I reverted that churn rather than folding it into this PR.

### DIFF
--- a/kinder-bilevel-planning/src/kinder_bilevel_planning/env_models/kinematic3d_v2/vega_pickplace3d.py
+++ b/kinder-bilevel-planning/src/kinder_bilevel_planning/env_models/kinematic3d_v2/vega_pickplace3d.py
@@ -43,13 +43,16 @@ def create_bilevel_planning_models(
     action_space: Space,
     config: VegaPickPlace3DEnvConfig | None = None,
     prefer_ompl: bool = True,
+    grasp_approach_max_tilt: float | None = None,
 ) -> SesameModels:
     """Create the env models for VegaPickPlace3D.
 
     ``config`` overrides the env config of the internal sim. Defaults to
     ``VegaPickPlace3DEnvConfig()`` if not provided. ``prefer_ompl`` selects the OMPL
     motion planner when ompl is installed; set it to False to force the BiRRT
-    fallback.
+    fallback. ``grasp_approach_max_tilt`` bounds how far (in radians) the end
+    effector may point off straight down in sampled pick and place configurations,
+    with the yaw free; None (the default) leaves grasp orientations unconstrained.
     """
     assert isinstance(observation_space, ObjectCentricBoxSpace)
     assert isinstance(action_space, BimanualArmJointDeltaGraspActionSpace)
@@ -139,7 +142,12 @@ def create_bilevel_planning_models(
     )
 
     # Controllers.
-    controllers = create_lifted_controllers(action_space, sim, prefer_ompl=prefer_ompl)
+    controllers = create_lifted_controllers(
+        action_space,
+        sim,
+        prefer_ompl=prefer_ompl,
+        grasp_approach_max_tilt=grasp_approach_max_tilt,
+    )
 
     # Finalize the skills.
     skills = {
--- a/kinder-bilevel-planning/tests/env_models/kinematic3d_v2/test_vega_pickplace3d.py
+++ b/kinder-bilevel-planning/tests/env_models/kinematic3d_v2/test_vega_pickplace3d.py
@@ -1,6 +1,7 @@
 """Tests for vega_pickplace3d.py."""
 
 import kinder
+import numpy as np
 import pytest
 from conftest import MAKE_VIDEOS
 from gymnasium.wrappers import RecordVideo
@@ -11,7 +12,7 @@ from kinder_bilevel_planning.env_models import create_bilevel_planning_models
 # VegaPickPlace3D needs prpl_kinematics, which is an optional extra of both kindergarden
 # and this package, and a kindergarden release that contains the environment. Skip the
 # module rather than fail collection when either is missing.
-pytest.importorskip("kinder.envs.kinematic3d_v2.vega_pickplace3d")
+vega_pickplace3d = pytest.importorskip("kinder.envs.kinematic3d_v2.vega_pickplace3d")
 pytest.importorskip(
     "kinder_models.kinematic3d_v2.vega_pickplace3d.parameterized_skills"
 )
@@ -58,6 +59,38 @@ def test_vega_pickplace3d_bilevel_planning():
         env = RecordVideo(env, "unit_test_videos", name_prefix="VegaPickPlace3D")
     # For seed 0 the cube and the target are both on the right side of the table.
     _run_bilevel_planning(env, seed=0)
+    env.close()
+
+
+def test_vega_pickplace3d_grasp_approach_max_tilt_plumbing():
+    """The tilt constraint threads through model creation to the pick sampler."""
+    max_tilt = 0.1
+    env = kinder.make("kinder/VegaPickPlace3D-v0")
+    env_models = create_bilevel_planning_models(
+        "vega_pickplace3d",
+        env.observation_space,
+        env.action_space,
+        prefer_ompl=False,
+        grasp_approach_max_tilt=max_tilt,
+    )
+    # For seed 0 the cube is on the right side of the table.
+    obs, _ = env.reset(seed=0)
+    state = env_models.observation_to_state(obs)
+    pick = next(s for s in env_models.skills if s.operator.name == "Pick")
+    arm = state.get_object_from_name("right_arm")
+    cube = state.get_object_from_name("cube")
+    params = pick.controller.ground((arm, cube)).sample_parameters(
+        state, np.random.default_rng(0)
+    )
+
+    sim = vega_pickplace3d.ObjectCentricVegaPickPlace3DEnv(allow_state_access=True)
+    sim.set_state(state)
+    sim.set_arm_joint_positions("right", params)
+    rotation = sim.end_effector_pose("right").R
+    tilt = np.arccos(np.clip(-rotation[2, 2], -1.0, 1.0))
+    assert tilt <= max_tilt
+
+    sim.close()
     env.close()
 
 
--- a/kinder-models/src/kinder_models/kinematic3d_v2/vega_pickplace3d/parameterized_skills.py
+++ b/kinder-models/src/kinder_models/kinematic3d_v2/vega_pickplace3d/parameterized_skills.py
@@ -12,6 +12,13 @@ grasp command to grasp, release, or take the cube.
 The handover skill moves both arms in sequence: the holding arm carries the cube into a
 region in front of the robot that both arms can reach, and the receiving arm then
 reaches to the cube and takes it.
+
+An optional grasp-approach tilt constraint (``grasp_approach_max_tilt``) restricts pick
+and place configurations to top-down approaches: the end effector points at most that
+angle off straight down, with the yaw free. When it is set, those two skills sample
+constraint-aware, drawing a yaw and solving full-pose IK for a down-facing target,
+because rejection over uniform draws essentially never produces a near-vertical end
+effector that also reaches the cube.
 """
 
 from __future__ import annotations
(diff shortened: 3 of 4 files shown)


tomsilver:
Verified against the prpl-dexmate real-table scene (cube at the hardware-calibrated grasp point): sampler-level, 8/8 pick+place samples within the 0.1 rad bound at 0.1-0.8 s per sample; end-to-end, a fake-mode rollout produced a right-arm-only plan whose grasp sits 4.1 deg off vertical and placed the cube inside the patch. Works as designed.

One hardware-motivated request before merge: some accepted samples land the wrist roll exactly at its soft limit (two of five pick seeds gave j7 = +1.38 vs the +/-1.378 limit). On the physical Vega, firmware silently zero-torques an axis that goes beyond its soft limit (dynamically, not just at boot), so commanding a grasp with j7 at the boundary risks a limp wrist on any tracking overshoot. Suggestion: have the down-facing samplers reject solutions within ~0.1 rad of a joint limit (or expose a margin), so constrained grasps stay servo-safe.

tomsilver:
Follow-up from testing this branch on the prpl-dexmate real-table scene: the sampled motions execute correctly but reconfigure the arm far more than the task needs, and the cause is in the down-facing samplers rather than the motion planner.

Measurements (staged cube at (0.46, -0.16, 0.647), target 15 cm away, OMPL RRTConnect with simplification): the planned paths are essentially straight in joint space (path-length to straight-line ratios 1.02 for pick, 1.08 for place), yet the place configuration lands **5.78 rad max joint delta** from the pick configuration. Moving the cube 15 cm costs a ~6 rad posture flip because pick and place each draw an independent yaw and IK seed, and the first valid solution wins with no preference for the posture the arm is already in.

Suggested fix for the place side, which is where the flip lives: when the tilt constraint is active, **do not resample a posture for the place at all**. The arm is already holding the cube in a valid down-facing grasp, so first try:

- target rotation = the **current end-effector rotation** (the pick yaw is as good as any other, and reusing it loses nothing since yaw is free), 
- target position = place point minus the rotated cube offset (as already computed here),
- IK **seeded from the current configuration**,

and fall back to the existing yaw-sampling loop only if that solve fails or collides. That makes the transfer a short translation at constant orientation instead of a reconfiguration, by construction rather than by heuristic.

The pick side has the mirror-image issue in milder form (home to pick is sampled with random seeds too); the principled fix there is an axis-alignment IK constraint (constrain the approach axis to vertical, leave rotation about it free) so the solver seeded from the current configuration finds the nearest-posture solution and the yaw falls out instead of being drawn — but that needs a partial-orientation mode in prpl_kinematics's NumericalIK, so it is probably a separate change.

tomsilver:
Both addressed in d4a4e7f.

**Soft limit margin.** Rather than rejecting boundary solutions after the fact, the constrained solvers now run in a margin-shrunken joint space (`_MarginJointSpace`, bounds pulled in by `DOWN_FACING_JOINT_LIMIT_MARGIN = 0.1`), so every accepted solution is at least 0.1 rad inside every soft limit by construction. Rejection-after-solve turned out to interact badly with the posture fix below: the seeded place solve would converge to a servo-unsafe boundary solution and get discarded, forcing a posture resample. Shrinking the space instead lets the solver converge to a nearby margin-respecting solution directly. The margin costs acceptance rate (hard case drops from 20% to 10% per candidate), so the candidate budget doubles to 100; a genuine exhaustion now takes ~25 s.

**Posture flip on place.** Implemented with one extension to your suggestion. The place sampler seeds every solve from the current configuration and starts at the current yaw, as you proposed. But reusing the current yaw alone was not enough: two of four trial seeds carried a wrist roll sitting exactly on the margin boundary (the same failure your j7 = +1.38 samples showed), and the constant-yaw solve stalls against that pin. And the nearest feasible *yaw* is not the nearest *posture* — unwinding a pinned wrist at a 27 deg yaw change cost 2.1 rad of roll, while a 62 deg yaw change cost 0.75 rad. So the sampler fans `PLACE_POSTURE_SEED_ATTEMPTS = 12` yaws out from the current yaw, solves each from the current seed, accepts immediately below 1.2 rad max joint change, and otherwise keeps the closest solution; full resampling remains the last resort (and the path taken when the held grasp is not down-facing, e.g. after a handover). One detail: the fan targets use the exactly-vertical rotation at each candidate yaw rather than the current rotation, so the converged tilt stays within the bound even when the held grasp is itself tilted.

Measured on sim seed 0 across four rng seeds (pick executed, then place sampled): max joint change from the pick posture went from 5.40 / 4.81 / 0.81 / 0.81 rad to 0.54 / 0.58 / 0.93 / 0.69 rad. The unit test now asserts posture preservation (< 1.5 rad) and the margin on both pick and place samples.

The pick side is untouched, per your comment — the axis-alignment partial-orientation mode in prpl_kinematics remains the principled fix there.

tomsilver:
Re-verified the updated branch on the prpl-dexmate real-table scene. The margin-shrunken joint space works (sampled confs sit >= 0.1 rad inside every limit), and the place fan logic is correct — but on our scene it still produced a 5.48 rad pick-to-place flip, and instrumenting the fan shows why: **all 13 seeded tiers fail, because the pick conf itself is the problem.**

The pick sampler (first-valid-wins over random seeds) returned `[0.89, 0.35, 2.97, -1.89, -1.88, -0.64, -0.34]` — j2 and j3 both pinned exactly at the 0.1 margin boundary. From a posture jammed against its bounds, the 15 cm transfer at fixed vertical orientation is unsolvable at any fanned yaw, so the fallback resamples and flips.

Control experiment: staging the held state at a well-conditioned down-facing pick conf (`[0.151, -0.783, -1.782, -1.72, -0.63, 1.392, 0.841]`, all joints comfortably inside limits) and rerunning the same place fan: **tier one succeeds on the first solve with a 0.37 rad transfer.** So the fan composes beautifully — it just needs the pick to feed it a healthy posture.

Suggestion: give the pick sampler the same preference the place fan has. Since there is no meaningful "current yaw" at pick time, the min-delta form fits better than the fan: keep drawing yaw/seed candidates as now, but instead of returning the first valid solution, collect valid ones and return the min max-joint-delta from the current configuration, with an early-accept threshold (the healthy conf above is 2.03 rad from home, so something like 2.5 rad would short-circuit the common case). Bonus: preferring postures near the current conf also biases away from margin-pinned solutions, since those tend to be the contorted branch. The axis-alignment IK constraint remains the principled long-term fix for pick, but min-delta selection should close the gap now.

tomsilver:
Pick side addressed in 81b011c, with one addition to the min-delta suggestion.

Before implementing I measured the delta-from-home distribution of valid pick solutions, and it surfaced something stronger than the delta signal: **over half of all valid down-facing solutions rest at least one joint exactly on the margin boundary** (18/30 easy case, 16/25 hard case) — the solver clamps to the shrunken space each iteration, so the boundary is sticky. Since a pinned joint is the direct cause of the place-fan stall (your j2/j3 case), the sampler scores each valid solution as `max_joint_delta_from_current + 10.0 * num_pinned_joints`, so an unpinned solution always outranks a pinned one and delta breaks ties. It early-accepts an unpinned solution under 2.5 rad (your suggested threshold; well-conditioned pick postures measure ~2 rad from home) and otherwise returns the best of `DOWN_FACING_SELECT_CANDIDATES = 5` valid solutions.

Results on sim seed 0 across four rng seeds, running pick then sampling place:

- Every pick sample is now unpinned (previously the pinned fraction among first-valid winners tracked the ~55% base rate).
- Pick-to-place max joint change: 0.63 / 0.84 / 0.81 / 0.79 rad, with the place fan's tier one succeeding immediately (place sampling ~0-0.5 s). Before the pick fix, two of these four trials needed the fan rescue and one still flipped at 5.4 rad in your scene's failure mode.
- Cost: easy-case pick sampling 0.6-1.4 s. The hard case (seed 19, left arm) ranges 3-18 s — the 12% early-accept rate there means the 5-valid window often fills at ~10% per-candidate acceptance. Slower than first-valid-wins (~2 s), but per the budget rationale in the module, sampling reliability and hardware quality outrank the sampling speed.

The selection lives in `_sample_down_facing_configuration`, so the place resampling fallback inherits the same preference. The unit test now asserts pick samples are strictly inside the margin boundary. The axis-alignment IK mode stays the long-term fix; happy to take that up in prpl_kinematics next if useful.

tomsilver:
Verified the pick-side update on the prpl-dexmate real-table scene — this closes the loop:

- End-to-end (fake pipeline, staged cube at the hardware grasp point): pick at 3.2 deg tilt with 0.48 rad limit margin, and the transfer is now **15 waypoints / 0.32 rad** (previously 180 waypoints / 5.48 rad). The whole motion stays at table level; the cube lands in the patch.
- Across 5 RNG seeds: pick-to-place deltas 0.32-0.43 rad, home-to-pick 2.2-3.1 rad (the inherent shoulder-forward reconfiguration), all confs >= 0.10 rad inside every joint limit.

From our side this is ready to merge. Once released we will bump pins and plumb `grasp_approach_max_tilt` through the prpl-dexmate pipeline config.

PR #137, opened 2026-08-19
Make Kinematic3D base navigation respect declared action bounds

merlerm:
## Summary

Make the Kinematic3D base-navigation skills respect the environment's declared action bounds:

- add the missing `remap_se2_pose_plan_to_constant_distance` call to the Shelf3D and PrplLab3D skills (the Transport3D skills already apply it), so BiRRT base plans are resampled to steps of at most `max_action_mag` per coordinate; and
- replace open-loop waypoint popping with a shared `step_toward_se2_waypoint` helper (in `kinder_models.kinematic3d.utils`) that clamps each emitted delta to the bound and only advances to the next waypoint once it is reached — so a clipped or otherwise short step is retried instead of silently skipping ahead and compounding drift. Applied to the shared `BasePlaceController` and the transport3d/shelf3d/prpl3d controllers.

## Motivation

Once kindergarden enforces the declared base-action bounds (Princeton-Robot-Planning-and-Learning/kindergarden#135), the old skills break on exactly the environments missing the remap: they emit base deltas up to 3× the bound (12–21 out-of-bound actions per Shelf3D episode, excess up to 0.43; up to 0.64 on PrplLab3D). Pre-enforcement, those actions executed oversized; post-enforcement they are clipped and the plans collapse (Shelf3D bilevel planning drops 0.81 → 0.21, PrplLab3D fails outright).

The 0.2 bounds on these environments were set deliberately in the prpl-mono collision-checking audit (#287); this change makes the skills honor them.

## Results

Full before/after evidence in the kindergarden PR thread (Princeton-Robot-Planning-and-Learning/kindergarden#135, validation comment). Headline, bilevel planning at 5 outer seeds × 50 episodes:

| Environment | old skills, pre-/post-enforcement | fixed skills, pre-/post-enforcement |
|---|---|---|
| Shelf3D-o2 | 0.81 / 0.21 | **1.000 / 1.000** (250/250 traces byte-identical) |
| PrplLab3D-o2 | 1.000 / hangs | **1.000 / 1.000** (250/250 traces byte-identical) |
| Transport3D-o2 | 0.284 / 0.280 | 0.300 / 0.300 (250/250 traces byte-identical) |

The fixed skills emit zero out-of-bound actions, are byte-for-byte identical across the enforcement change, and outperform the old skills even on pre-enforcement kindergarden. (Note: full Shelf3D success additionally requires the planning-config change in the companion PR — the compliant skills take more, smaller steps per navigation, so `max_skill_horizon` must be raised as it already is for Transport3D.)

## Test plan

- `kinder-models` transport3d + shelf3d skill suites: **6/6 pass** (and pass unchanged on the pre-fix code, i.e., the suites are compatible with both behaviors for in-bound plans).
- 1,750 paired regression episodes across bilevel planning and the scripted experts: zero out-of-bound actions, zero behavioral change on pre-enforcement kindergarden beyond the removed violations.


kinder-models/src/kinder_models/kinematic3d/utils.py
```diff
@@ -23,3 +23,25 @@ def get_target_robot_pose_from_parameters(
 
     # Robot faces the target: heading points along +ang (toward the target).
     return SE2Pose(rx, ry, ang)
+
+
+def step_toward_se2_waypoint(
```
yixuanhuang98, line 28: Could we have a unit test for this function?

yixuanhuang98 (approved):
LGTM! Left one comment.

PR #140, opened 2026-08-19
Persist environment normalization statistics in RL checkpoints

merlerm:
## Summary

Make RL checkpoints self-contained: PPO/SAC train inside `NormalizeObservation`/`NormalizeReward`, whose running statistics live in the env wrappers — but `agent.save()` persisted only network and optimizer state, so a loaded checkpoint fed the policy observations on the wrong scale and could not reproduce the trained agent. This PR:

- captures the per-env `obs_rms`/`return_rms` statistics at the end of `train()` (before the envs close), for both PPO and SAC, with a log line stating what was captured;
- includes them in the checkpoint as plain lists/floats, keeping `torch.load(weights_only=True)` compatibility (torch ≥2.6 default rejects pickled numpy arrays);
- restores the field on `load()`; and
- adds `set_env_normalization_stats` to apply them to synchronous evaluation envs (explicit `NotImplementedError` for async, where `get_attr` returns copies and mutation would be silently lost).

Legacy checkpoints load unchanged (`env_normalization_stats=None`).

## Motivation

Found while validating kindergarden#135: we intended to evaluate the released frozen checkpoints before/after the environment change, but the released `agent.pkl` archives cannot reproduce the trained policies — the normalization statistics they depend on were never saved — forcing full retraining instead. Any future frozen-checkpoint evaluation, checkpoint release, or resume-from-checkpoint workflow hits the same wall.

## Test plan

- Train → save → `torch.load(weights_only=True)` succeeds and contains per-env stats with the expected counts.
- Fresh-process `load()` → `set_env_normalization_stats` into a new `SyncVectorEnv` → statistics round-trip exactly (means/vars/counts bit-equal).
- SAC path verified through `AsyncVectorEnv` capture.
- Backward compatibility: a checkpoint without the new key loads with `stats=None`.
- A behavioral A/B (frozen eval with vs without restored statistics on a fully trained BaseMotion3D PPO agent) is running; results to follow as a comment.

### DIFF
--- a/kinder-rl/src/kinder_rl/gym_utils.py
+++ b/kinder-rl/src/kinder_rl/gym_utils.py
@@ -425,3 +425,52 @@ class ReplayBuffer(BaseBuffer):
         if dtype == np.float64:
             return np.float32
         return dtype
+
+
+def get_env_normalization_stats(envs: gym.vector.VectorEnv) -> list[dict]:
+    """Collect the NormalizeObservation/NormalizeReward running statistics.
+
+    Returns one entry per sub-environment. Works for both Sync and Async vector
+    envs (Async returns copies, which is sufficient for checkpointing).
+    """
+    obs_rms = envs.get_attr("obs_rms")
+    return_rms = envs.get_attr("return_rms")
+    # Plain lists/floats keep checkpoints loadable under torch.load's
+    # weights_only=True default (torch >= 2.6 rejects pickled numpy arrays).
+    return [
+        {
+            "obs_mean": np.asarray(o.mean).tolist(),
+            "obs_var": np.asarray(o.var).tolist(),
+            "obs_count": float(o.count),
+            "return_mean": np.asarray(r.mean).tolist(),
+            "return_var": np.asarray(r.var).tolist(),
+            "return_count": float(r.count),
+        }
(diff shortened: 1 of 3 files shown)


tomsilver (approved):
You're a PR machine!

merlerm:
Thank Claude for finding 200 crazy edge cases when all it had to do was run some baselines hahah, but these were all things it fixed when it was running them, so they were all ready, just had to check them and clean up. Will merge everything tomorrow though 😴

merlerm:
Behavioral A/B results, as promised in the description — PPO BaseMotion3D trained to completion with this branch (1M steps, released configuration, seed 301), then frozen-evaluated from the checkpoint in a fresh process (50 episodes, same eval seed):

| Frozen eval | Mean episodic return |
|---|---:|
| **with restored normalization statistics** | **−88.6** |
| without them (fresh `NormalizeObservation` state) | −100.0 (failure floor — 0/50 episodes reach the goal) |
| training-time eval anchor | ≈ −98 (released −98.5) |

With the statistics restored, the frozen checkpoint performs in the trained regime; without them the same network is completely non-functional — the observations arrive on the wrong scale and the policy never solves a single episode. This is the measured version of the PR's premise: checkpoints without these statistics cannot reproduce the trained agent at all.

PR #144, opened 2026-08-23
kinder_models: PickCube base-motion collision, multi-rotation grasp, and arm-stall fixes

joshnroy:
## Summary
- ~~Splits `END_EFFECTOR_TO_OBJECT_HOLDING_TOLERANCE` into separate XY (0.02) and Z
  (0.08) tolerances for `Tossing3DStateAbstractor._check_holding` -- one shared
  tolerance made the real, roughly-constant vertical TCP-to-cube-center grasp offset a
  frequent near-miss (measured dz=0.0610 against a 0.0600 tolerance, dx/dy both <0.01
  the same tick). The original shared constant stays at 0.06 for its other callers
  (shelf, motion-planning).~~ **Reverted (2026-08-24 update, below): A/B sweeps found
  this didn't earn its complexity.**
- ~~Fixes `HandEmpty` to also gate on "not holding anything," not just "gripper
  commanded open" -- previously a gripper that closed on nothing could read as neither
  `HandEmpty` nor `Holding`.~~ **Corrected below (2026-08-24 update): this was a no-op
  and has been reverted.**
- Fixes `PickCubeController.reset()`'s `BASE_MOTION` planning call to exclude the
  target cube from collision (`disable_collision_objects=[cube_to_pick_up.name]`),
  matching what the toss controller's own base-motion call already does and for the
  identical documented reason. Without it, a robot that ends up parked within ~0.3m of
  the cube (inside its own 0.5x0.5m footprint) has its own start pose in collision, and
  `BASE_MOTION` planning fails deterministically every time until something else
  relocates the robot -- the root cause of an intermittent "controller does nothing"
  symptom late in a practice run.
- Adds a `RETURN_HOME` phase to `MoveToTossLocationAndTossController` that drives the
  arm to a fixed stowed configuration after the swing, for readable post-toss video.
- Replaces ad hoc diagnostic monkeypatches used while chasing the above with permanent
  `logger.debug()` calls at the phase/plan level.

## Test plan
- `pytest` against the consuming project's Tossing3D suite (202 passed, 1 xfailed)
  after these changes plus the operator-model changes on the consuming side.
- Manually verified via repeated single-episode repro runs (practice + eval) that the
  base-motion collision fix eliminates the deterministic planning-failure streak, and
  that eval episodes now dispatch real skills rather than stalling out.

## Followup
- The physically-correct `Holding` check (finger-contact query rather than a
  position-delta threshold) is blocked on `PyBulletSim.set_state()` hardcoding the
  gripper/finger joints to 0.0 -- unrelated to the real `pos_gripper` command, so no
  contact query against it can see a real grasp. Fixing that needs real per-joint
  gripper state restored on `set_state()` (see PR #162/#164, unmerged).

---

**Update:** adds a third commit -- `PickCubeController` now tries all four
grasp-symmetric rotations for `BASE_MOTION` planning instead of hardcoding the
nearest-measured-yaw one. Found via a 10-seed, 50-cycle verification sweep: 8
of 9 finished seeds hit systematic `BASE_MOTION` planning failure within the
first few cycles once a missed toss left the cube near the barrier, and stayed
stuck for nearly the rest of the run (48-49 of 50 cycles reduced to near-zero
physical content). Root cause: the standoff angle is exactly the cube's own
measured yaw with zero adjustment (`TARGET_ROTATION=0.0`), and only the single
nearest-yaw grasp rotation was ever tried, even though the cube is genuinely
four-fold symmetric for grasping and `upright_grasp_rotations`' own docstring
anticipates a caller falling through the other three. Verified against the
exact stuck state from the sweep (full details in the commit message).

---

**Update (2026-08-24):** adds a fourth commit -- a cleanup pass, no behavior change
except one correction:

- Trims oversized comments/docstrings to their load-bearing WHY, and removes
  now-redundant per-tick/per-transition `logger.debug()` calls, across the
  grasp-symmetric `BASE_MOTION` fix, the stall-replan watchdog, `RETURN_HOME`, and
  the per-axis holding tolerance.
- **Reverts the `HandEmpty` change above.** `_check_holding`'s own gate
  (`pos_gripper > GRIPPER_GRASPING_THRESHOLD`) and `_check_gripper_open`'s gate
  (`pos_gripper ~= 0`) are mutually exclusive on `pos_gripper`, so whenever
  `_check_gripper_open` is `True`, `_check_holding` is already forced `False` by its
  own gate -- the added `and not any(holding.values())` clause never changed the
  result. `HandEmpty` is back to `self._check_gripper_open(state, robot)` alone.
- `PickCube`'s multi-rotation grasp approach, its base-motion collision exclusion,
  and `RETURN_HOME` are all kept as-is.

Also ran an A/B sweep on the multi-rotation grasp fix (grasp-symmetric
`BASE_MOTION`, above) to check whether it's earning its complexity: a controlled
630-point grid (3 robot start poses x 21 barrier clearances 0-0.5m x 10 cube yaws
0-90 deg) comparing OLD (nearest-yaw-only) against NEW (try all four
`upright_grasp_rotations`). OLD fails 84/630 (13.3%); NEW fails 0/630 (0.0%) --
every OLD failure is rescued by the fix, none remain unrescuable. Since the loop
breaks on first success, there's no added cost in the common case where
nearest-yaw already works, so the fix stays unconditional (already what
`PickCubeController.reset()` does). One caveat found while rendering, not fixed by
or in scope for this fix: at the very tightest clearance (~0.02m) the pick doesn't
reliably complete within the step budget even with the fix, though `BASE_MOTION`
planning itself succeeds -- looks like a separate, uncollision-checked arm-vs-barrier
issue during the reach/grasp phase.

Test plan: `pytest tests/environments/tossing3d` (hitl-pmp, consuming project)
still 202 passed, 1 xfailed -- unchanged. black/isort on the touched files;
pylint 10.00/10.

---

**Update (2026-08-24, second):** three more commits.

- Extracts `PickCubeController.reset()`'s grasp-symmetric `BASE_MOTION` loop into
  its own method, `_plan_grasp_symmetric_base_motion`. No behavior change.
- Adds a `NoOpController` tossing skill (single-tick, does nothing -- zero
  base/arm motion, and re-issues the robot's own current gripper command rather
  than zeroing it, since zero means "open" and would drop anything held).
  Registered as `"no_op"` alongside the other single-object skills.
- A round of independent code review on the diff turned up one real bug and one
  cleanup: `MoveToTossLocationAndTossController._action_return_home` was
  P-controlling toward its fixed retract target with a raw `target - curr`
  instead of `wrap_arm_joint_difference(...)` -- the retract target holds joint
  index 2 at 180 degrees (the wrap discontinuity itself), and the arm is expected
  to be sitting right at that boundary when `RETURN_HOME` starts (windup/swing/
  release hold the same joint there), so a raw subtraction there could report
  ~2*pi of travel between two identical poses and send the arm the long way
  around a joint. Fixed to match `_step_trajectory_phase`'s existing pattern.
  Also dropped a leftover duplicate `robot = self.objects[0]` in `reset()` the
  extraction above left dead.
- **The same review also caught that `END_EFFECTOR_TO_OBJECT_HOLDING_TOLERANCE`
  had silently changed value (0.05 -> 0.06) in an earlier commit on this branch,
  not just been "kept" as the first Summary bullet implied, and that Shelf's own
  `Holding` predicate depends on the same shared constant** -- prompting two more
  A/B sweeps, both real full-physics `PickCube` grasps across an 81-point barrier
  clearance x cube yaw grid: the shipped XY/Z split (0.02/0.08) against a single
  0.06 (of 63 real grasps, single-0.06 called 62/63 Holding, the split only
  56/63 -- 6 new false negatives at cube yaw~50 degrees from a stable ~3cm
  x-offset the split's tight 0.02 xy bound rejected, 0 rescues), and single-0.05
  against single-0.06 on the same 63-point data (zero disagreements -- no real
  grasp ever landed between the two values on any axis, closest margin 1.5cm).
  Neither the split nor the earlier 0.05->0.06 bump earned its complexity, so
  `_check_holding` is reverted to exactly `main`'s original single
  `END_EFFECTOR_TO_OBJECT_HOLDING_TOLERANCE = 0.05` implementation. The one real
  finding from the sweeps -- a stable false-negative cluster at cube yaw~50
  degrees -- isn't fixed by any tolerance value tested (it would need a *wider*
  xy bound, the opposite direction from where the split moved) and is out of
  scope here.

Test plan: kinder-baselines' own `pytest tests/dynamic3d/shelf tests/dynamic3d/tossing`
(this repo, DISPLAY set to sidestep an unrelated sandbox-only rendering-backend
quirk): 54 passed, both before and after the tolerance revert. `pytest
tests/environments/tossing3d` (hitl-pmp, consuming project): 202 passed, 1
xfailed, unchanged across every commit above. black/isort on touched files;
pylint 10.00/10.


joshnroy:
Superseded by a stacked split: #145, #146, #147, #148, #149 (same content, split into independent, reviewable pieces). Closing this one in favor of the stack.

PR #147, opened 2026-08-25
kinder_models: add a stall-replan watchdog to PickCube's trajectory phases

joshnroy:
## TL;DR
Adds a stall-replan watchdog to `PickCube`'s trajectory-following phases: if the waypoint index hasn't advanced for `STALL_REPLAN_PATIENCE` (40) ticks, replan the phase's remaining path from the arm's live conf.

## Question / goal
Why does a `PickCube` dispatch sometimes burn its entire step budget with zero progress, stuck at the same waypoint?

## Background
`_step_trajectory_phase` drives the arm through a pre-computed, collision-checked path one waypoint at a time. One observed dispatch had `LIFT_CUBE_TO_HOME`'s waypoint index frozen at 0/26 for 276 ticks straight -- the low-level P-controller fighting something (contact, an awkward IK solution) it never resolves on its own -- with no mechanism to detect or recover from this.

## Hypothesis
Detecting "no progress for N ticks" and replanning from the arm's actual live configuration (rather than the original plan's assumed trajectory) recovers dispatches that would otherwise stall for their entire budget.

## Guidance given
Built and evaluated as a comparison against wiring in a separate, existing-but-unused arm-settling gate; the two are alternate fixes for related symptoms, and this one is the one that shipped.

## Methods
Tracks consecutive ticks without waypoint-index advancement (any waypoint, not just the phase's final one) per phase in `_stall_ticks`/`_prev_plan_idx`. Past `STALL_REPLAN_PATIENCE` (40) ticks stalled, calls `_replan_phase_from_current_state` to re-run motion planning from the arm's live conf to the same final target the original plan aimed for. If the live conf is genuinely infeasible to replan from, the stalled plan is left in place and the watchdog retries after another `STALL_REPLAN_PATIENCE` ticks, rather than looping the (expensive) replan attempt every tick.

## Results
Verified as part of the larger branch this was originally developed on (now split into this stack): `pytest tests/environments/tossing3d` (hitl-pmp, consuming project) 202 passed, 1 xfailed; kinder-baselines' own `tests/dynamic3d/{shelf,tossing}` 54 passed. black/isort; pylint 10.00/10.

## Recommendation
Land as a reliability fix for a real, observed stuck state.

### DIFF
--- a/kinder-models/src/kinder_models/dynamic3d/tossing/parameterized_skills.py
+++ b/kinder-models/src/kinder_models/dynamic3d/tossing/parameterized_skills.py
@@ -83,6 +83,10 @@ ARM_SETTLE_TOLERANCE = 5e-3
 # waypoints: handing off to the gripper at 4e-2 leaves the grasp ~2.8cm off the cube.
 ARM_FINAL_CONF_TOLERANCE = 5e-3
 
+# Ticks a trajectory phase may sit at its final waypoint without completing before
+# the stall watchdog replans its remaining path from the arm's live conf.
+STALL_REPLAN_PATIENCE = 40
+
 
 class MoveToTargetGroundController(
     GroundParameterizedController[ObjectCentricState, Array]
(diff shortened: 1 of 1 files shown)


joshnroy:
Closing for now, not shipping yet. A real A/B (one seed, 50 practice cycles, same seed both arms) confirms the fix does what it claims: it doesn't reduce stall frequency (168 stalls without the watchdog vs 167 with -- same underlying rate), but it caps the worst-case cost. Without the watchdog, 2/168 stalled dispatches blow up to 337 ticks (the same kind of runaway the PR's own motivating case describes); with it, every one of 167 stalls stays capped at <=215 ticks, with zero infeasible-replan failures. PickCube itself never outright failed in either arm (52/52 vs 50/50).

The reason to hold off: final eval success swung a lot between arms (90% without, 40% with) on this single seed. Skill-level metrics look equivalent (MoveToTossLocationAndToss ~16-17% success both arms), and the likely mechanism is RNG cascade -- the watchdog's first replan changes that dispatch's exact tick-by-tick trajectory, which cascades into different downstream sampler draws for the rest of the 50-cycle run -- not a watchdog regression. But the eval curve shapes differ more than I'd like to wave away: WITHOUT converges to a stable 8-10/10 plateau by cycle 15; WITH oscillates 1-10/10 for the entire run with no clear convergence. That's not proof of a real effect on n=1, but it's enough that I want to think about it more before it goes to review, possibly with a multi-seed check.

Coming back to this. Branch (kb144-split-3-stall-replan) stays as-is in the meantime.

PR #149, opened 2026-08-25
kinder_models: add a NoOpController tossing skill

joshnroy:
## TL;DR
Adds a `NoOpController` tossing skill: a single-tick controller that does nothing, for use wherever the planner/practice loop needs a genuine no-op dispatch.

## Question / goal
None -- implementation task, not an experiment.

## Background
The tossing skill set (`PickCube`, `MoveToTossLocationAndToss`, `OpenGripper`, `CloseGripper`, `move_to_target`, ...) has no skill that dispatches and genuinely does nothing.

## Hypothesis
None -- implementation task, not an experiment.

## Guidance given
Requested directly, to have a genuine no-op skill available in the skill set.

## Methods
A single-tick controller: zero base/arm motion, and re-issues the robot's own current gripper command rather than zeroing it -- a bare all-zero action would command the gripper OPEN (see `OpenGripperController.step()`), which is not a no-op if the robot is currently holding something. Registered in `create_lifted_controllers` as `"no_op"`, alongside the other single-object skills.

## Results
New test: terminates immediately, all-zero action with the gripper index open (0.0), and with the gripper index preserving `GRASP_CLOSE_THRESHOLD` when the robot is holding. `pytest tests/environments/tossing3d` (hitl-pmp, consuming project) 202 passed, 1 xfailed; kinder-baselines' own `tests/dynamic3d/{shelf,tossing}` 54 passed. black/isort; pylint 10.00/10.

## Recommendation
Land as a straightforward new capability.

### DIFF
--- a/kinder-models/src/kinder_models/dynamic3d/tossing/parameterized_skills.py
+++ b/kinder-models/src/kinder_models/dynamic3d/tossing/parameterized_skills.py
@@ -759,6 +759,49 @@ class OpenGripperController(GroundParameterizedController[ObjectCentricState, Ar
         return current_gripper_pose < atol
 
 
+class NoOpController(GroundParameterizedController[ObjectCentricState, Array]):
+    """Controller that does nothing for a single control step.
+
+    The object parameters are:
+        robot: The robot itself.
+
+    Re-issues the robot's own current gripper command rather than zeroing it --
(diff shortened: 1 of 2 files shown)


yixuanhuang98 (approved):
LGTM

PR #155, opened 2026-08-27
Add a top-down grasp mode behind a flag; side grasp stays the default

tomsilver:
## Summary

Adds a top-down grasp option to the CylinderShelf3D skills behind a `grasp_mode` flag ("side" remains the default and is behaviorally unchanged). Both modes are canned planar motions through joints 2/4/6 with the joint-1 lateral correction; `top_down` descends vertically onto the cylinder and grips its upper section (3 cm below the top), sampling the approach angle over the full circle like the side grasp.

Two findings baked into the implementation:

- **The vertical reach needs a *farther* base, not closer**: with joints 3/5/7 pinned at retract, the wrist cannot point straight down near the body, so the top-down pick stages at 0.68–0.78 m (vs 0.78–0.88 m for side).
- **Top-down placement is infeasible on the standard shelf**: the place must enter at the grasp's pitch (the only way to keep the cylinder upright), and in pitch-90 planar configurations the gripper-and-forearm column above the held cylinder measures 0.61–0.66 m — more than any finite shelf opening. The mode-aware opening fit check fails fast there, so `top_down` currently supports pick-only experiments on this shelf (and placement onto open-topped surfaces in future envs). The place also gains a degenerate-yaw guard for vertical approaches and a horizontal pre-place backoff in top-down mode.

## Testing

- New tests: top-down pick succeeds (planar joints still frozen), and top-down place on the standard shelf raises `TrajectorySamplingFailure` every sample; existing side-mode tests unchanged and passing.
- mypy / pylint clean on the changed files.

Companion PR in prpl-tidybot exposes `env.grasp_mode`.

### DIFF
--- a/kinder-bilevel-planning/src/kinder_bilevel_planning/env_models/kinematic3d/cylinder_shelf3d.py
+++ b/kinder-bilevel-planning/src/kinder_bilevel_planning/env_models/kinematic3d/cylinder_shelf3d.py
@@ -43,6 +43,7 @@ def create_bilevel_planning_models(
     action_space: Space,
     num_objects: int = 1,
     config: CylinderShelf3DEnvConfig | None = None,
+    grasp_mode: str = "side",
 ) -> SesameModels:
     """Create the env models for cylinder shelf 3D.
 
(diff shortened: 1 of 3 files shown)


tomsilver:
Closing: the top-down grasp mode is not being pursued. The magic-skills work (#156, #157) has been rebased onto main and no longer depends on this branch.

PR #170, opened 2026-09-14
kinder-pddlstream-planning: add the LimbRepositioning3D baseline

Minseo10:
## Summary

New baseline in `kinder-pddlstream-planning`: using LimbRepositioning3D from kindergarden.

## Changes

- `domain.pddl` / `stream.pddl`: three actions, four generator streams (`sample-grasp`, `plan-grasp-motion`, `plan-base-motion`, `plan-limb-motion`), and three test streams (`check-human-joint-limits`, `check-human-torque-limits`, `check-robot-torque-limits`).
- `stream.py`:
   - `sample_grasp`: grasps sampled at slides along the limb, each way up
   - `plan_grasp_motion`: compute a base pose, and the arm path from selected grasp. A base pose is kept only if the arm can hold the grasp at every point on the limb's way from start to goal, and the base is collision-free. Poses where the arm cannot hold the limb's weight are tried last.
   - `plan_base_motion`: a collision-free base path between two base poses.
   - `plan_limb_motion`:  runs closed-loop predictive-sampling MPC (`PredictiveSamplingMPC`) with rollouts in the environment itself, rejecting rollouts that break a limit.
   - `check_human_joint_limits`, `check_human_torque_limits`, `check_robot_torque_limits`: reject a trajectory that leaves the person's range of motion, overloads their joints, or exceeds the robot's torque limits.
   - `LimbStreamContext`: state shared by the streams. Treats the person as an obstacle for the base and arm, but not the grasped limb.
- `utils.py`: coupled robot-limb state capture/restore, the grasp weld, gravity and hold torque, the torque on the person's joints, and collision helpers.
- `run.py`: solves one of sixteen variants (`--variant`) or all of them (`--all-variants`) with the `adaptive` algorithm. A failed run replays its closest attempt into the GIF.

## Notes

- Supersedes #100.
- Needs `kindergarden>=0.2.5` for the limb RoM and body mass models and the limb resting on the furniture.
- `bed-*-leg` and `wheelchair-*-leg` skip base collision checks through per-variant overrides in `run.py`, since their goal is mostly reachable with the base near the furniture in this setup.

## Tests

- `test_run_limbrepositioning3d.py` added: base placement, grasping, collisions with the person, MPC progress, the three safety limits, and state restore.
- `pytest tests/`: 35 passed. mypy and pylint clean.


tomsilver (approved):
Nice work!

